PR #7, opened 2019-03-29
Accordion Adjustments

Qakulukiam:
- ~~added field for hardcoded aria label~~ removed aria label
- refactored some code
- added missing escaping
- removed redundant elements
- added inner div for content with comments in stylus to prevent coding mistakes I've seen multiple times before

### DIFF
--- a/Components/AccordionDefault/index.twig
+++ b/Components/AccordionDefault/index.twig
@@ -8,7 +8,7 @@
               {{ panel.panelTitle|e }}
             </button>
             <div class="panel-content" aria-hidden="true" id="content-{{ loop.index }}">
-              <div class="content">
+              <div class="panel-inner">
                 {{ panel.panelContent|e('wp_kses_post') }}
               </div>
             </div>
--- a/Components/AccordionDefault/index.twig
+++ b/Components/AccordionDefault/index.twig
@@ -1,7 +1,7 @@
 <div class="flyntComponent" is="flynt-accordion-default">
   <div class="container">
     <div class="content">
-      <ul class="accordion" aria-label="Accordion Panel">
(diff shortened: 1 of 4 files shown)


Components/AccordionDefault/index.twig
```diff
@@ -1,15 +1,16 @@
 <div class="flyntComponent" is="flynt-accordion-default">
   <div class="container">
     <div class="content">
-      <ul class="accordion" aria-label="Accordion Panel">
+      <ul class="accordion" aria-label="{{ ariaLabel|e }}">
```
steffenbew: Do we actually need this aria label at all? I don't get the accessibility value on this one and found w3 doesn't even have it in their example: https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html

Qakulukiam: True, I don't really see the immediate value of this label, though I didn't test on a screen reader yet. If we're talking accessibility, this seems to also be a good summary resource for accordions: https://davatron5000.github.io/a11y-nutrition-cards/components/accordion

steffenbew: Let's remove it then. Makes the component and content maintainability simpler. If we find out we need it (or another aria-label), I would be happy to add it in again.

Qakulukiam: I believe this is where the aria label might have originated: http://web-accessibility.carnegiemuseums.org/code/accordions/. They don't provide a reason other than "to let the user know explicitly that they are dealing with accordions" and their source U.S. Web Design System listed below the article also doesn't add an aria-label.

I removed the aria-label in the latest commit.

PR #9, opened 2019-04-04
Stylelint Adjustments

Qakulukiam:
- use [stylelint-config-sass-guidelines](https://www.npmjs.com/package/stylelint-config-sass-guidelines) as stylelint base config
- remove some rules (nesting depth, max compound nesting)
- fix all new linting errors
- rename all `style.scss` files in Components to the more appropriate and consistent `_style.scss`, because only scss files that are to be compiled to a CSS file should be without `_` in front
- add some stylelint ignore comments (due to problems with slick slider and scss globbing plugin)
- add single quotes manually to `[is=flynt-some-component]` (should be `[is='flynt-some-component']`), linter doesn't seem to catch this as it's valid CSS, but should not be a best practice [due to reasons](https://mathiasbynens.be/notes/unquoted-attribute-values). Using single quotes as opposed to double quotes, because that is [best practice according to SASS Guidelines](https://sass-guidelin.es/#quotes).


steffenbew:
Thanks, appreciate your updates, especially in regards to linting errors and selector best practices.
I have some remarks though:
1) I personally like the `nesting depth` and `max compound nesting` rules to prevent deeply nested selectors. In which cases are you / others running into issues with those?
2) I don't like the `_` for the `style.scss`. I get your point and it is correct according to documentation, but to my interpretation, a component opens up its own scope and is not just a partial include of the website, although it _is_ getting loaded as a partial into the `main.scss`.

Qakulukiam:
Thanks for your feedback!

1a. We have run into many issues with limiting nesting depth before. This happens especially with plugins that have very specific styling, which we need to override sometimes. But it also happens with more complex components. We could introduce a larger `nesting depth` setting of maybe around 3, but this will most likely lead to workarounds that I've seen before like *removing the setting from the linting rules*, *setting it to a higher value* or even *ignoring whole style files* because of it. 

1b. The other setting for `max compound nesting` (set to 3 by default) is probably an even bigger issue, because - in addition to the issues pointed out for the other setting - of the way partials are structured and imported (as nested dependencies). We could also increase this value to maybe around 5-6 to at least have our approach work in general, but still have some way of enforcing best practices. If we make this adjustment, we should re-evaluate the setting after building a few production websites with it.

I'm open to introducing both of these linter settings with higher values than the defaults, we just need to monitor and be aware of the issues we could run into with them. Let me know if that's something you would like to try.

2. I don't really see the `_` as a scoping mechanism in sass, especially since you can write anything into these files (see NavigationMain for the basic example of adding styles to the html element; other examples I've seen before are template related themings in component styles). One large benefit for me would be that it's obvious to any sass developer how these files are bundled and created without a need to explain anything. If I see a `style.scss` I would presume that the file will be compiled as a standalone file. For a `_style.scss` I would not make this presumption.

Let me know if `style.scss` is your preferred solution anyways and I'll undo the change.

Qakulukiam:
We discussed this further internally and have come to the following agreement:

- The `max-nesting-depth` setting will be left disabled as we could not find a good solution without changing our styling approach. Main reasons for necessary nesting are the `&` parent reference feature and media queries. As this can easily lead to frustration, we have left this disabled.
- We have set the `selector-max-compound-selectors` to 4, which is one above the sass guidelines default. We deem this increase necessary in order to make up for our scoping approach using `[is='flynt-component-name']`. We believe including this linting feature will overall lead to improved code quality and should reduce specificity problems arising from complex code.
- We have decided to keep the commonly accepted `_style.scss` naming for component style files in order to make it easier for SASS developers to understand that a component file is not a standalone file, but one that is imported into the bundle. It further helps make clear that no variables need to be imported into the files as these are included in the `main.scss` and `admin.scss` respectively. This should, in our expectation, lead to less confusion when working with components.

PR #10, opened 2019-04-09
Add more flexibility to mixins

emcarru:


emcarru:
> @emcarru Should we also rename `list-reset` to `reset-list` according to `reset-content-margins`?

Yes! I've just pushed this change.

emcarru:
I've just looked back at this and remembered `display: block` is missing from `center`. I believe we briefly discussed this but to me it makes more sense to add `display: block` since otherwise `center` doesn't actually always center whatever you call it on. Thoughts @steffenbew @Qakulukiam ?

steffenbew:
@emcarru sure, add it.

emcarru:
> @emcarru sure, add it.
Done!

Qakulukiam:
@emcarru @steffenbew could one of you tell me why `position: relative` is in the `center` mixin? If we don't need it, I would prefer removing it..

steffenbew:
@Qakulukiam opening up a new positioning context with a new wrapping container is a best practice to me. It's not necessary for the desired centering, but not having position relative defined on such an element usually leads to unexpected behavior and tough debugging later, in my experience. 
On the flipside, it's adding this tiny bit of "magic behavior", whereas I would like to keep default code as obvious and as easy to understand as possible.

Qakulukiam:
@steffenbew I think the best practice argument could go both ways: For me it's unexpected that centering something leads to it being a new context. Especially with navigations I can think of a few use cases where this would confuse me if I didn't know about it and make me reset the styles. My vote is to remove this bit of magic and keep these concerns separate.

steffenbew:
@emcarru tie-break. your call.

emcarru:
I'm in favour of less 'magic' - lets remove it for now.

PR #16, opened 2019-05-09
fix(_base.scss): remove overflow setting on <html>

Qakulukiam:
This keeps leading to issues with plugins like body scroll lock and should not be needed on the html element anyways since it's already set on the body.


steffenbew:
I remember it not working only on the body element, but specifically needed to be on both, html and body element, to work on iOS Safari.

Qakulukiam:
outdated, fixed in master

PR #18, opened 2019-05-09
feat(customPostTypes/example): add revisions to 'supports'

emcarru:
Just a small thing - I used the example when I wanted to create a new post type and spotted that revisions are not there by default. I think having it enabled is a sensible default - better that you have this enabled and don't need it, rather than having it disabled when you later realise you need it and forgot.


Qakulukiam:
@emcarru I literally just copied the basic example from https://generatewp.com/post-type/ for this and adjusted syntax for Flynt, because I didn't want to get into discussions about the perfect example file. For me this file is a stepping stone anyways and I would much prefer to have this functionality in the Flynt Generator or Snippets instead.

If that's too slow for you, feel free to merge.

Issue #17, opened 2019-05-09
Optional Gutenberg support


bdbch:
Hey FlyntWP / bleech team! Hope everything is fine!

For the last 4 days I played around with the new ACF Pro Gutenberg Blocks in ACF Pro 5.8 and I think it's now way easier to register Flynt components into Gutenberg and also load necessary style and javascript files for the admin area.

I worked on a prototype and got dynamic component registration working for Gutenberg outside of the Flynt ecosystem. I started to implement optional Gutenberg component registration and Component rendering into the Flynt theme yesterday and look forward to send a pull request very soon.

Are you already working on something similar?

My current state looks as following:

1. Implemented a `useGutenberg` protected static variable in the `Defaults` class
2. Added the useGutenberg static function to update the protected variable
3. Added `useGutenberg(false)` into the Init.php
4. Added a check for the useGutenberg variable in the RemoveEditor.php file (If set to true, the default editor will be loaded with Gutenberg activated)
5. Disabled Component Registration via ACF layouts in the API class when useGutenberg is true

Now I'm working on the registration via [acf_register_block_type()](https://www.advancedcustomfields.com/resources/acf_register_block_type/) and updating the renderer to use the_content() if Gutenberg is enabled instead of the dynamic renderer via the ACF layout field.

Obviously there are some differences and incompatibilities with the ACF layout approach and the Gutenberg approach.

I also think there should be an option to remove the default Wordpress blocks to keep Flynt components clean and simple and avoid mixing them with default Wordpress blocks.

Happy to hear from you and have some insights or thoughts from you.

PS:
The new and simpler approach back to the Wordpress roots is really refreshing and fun to work with! Great job!

bdbch:
I'm stuck right now at rendering the output. Registration of the block and field group was successful as seen in the following screenshot:

![image](https://user-images.githubusercontent.com/6538827/57478199-a4631e00-729a-11e9-8176-81e2250dddae.png)

I can also edit the data and show fields in the gutenberg editor.

![image](https://user-images.githubusercontent.com/6538827/57478241-bd6bcf00-729a-11e9-98fb-0f05b8473a8a.png)

I'm using a var_dump of the passed data as the renderCallback right now so I can see data-passing actually works. The problem now is rendering the output via Twig. It looks like the Twig-Environment and correct renderComponent function is inside the `TwigExtensionFlynt.php` file with now way to render it directly.

[...]

PR #19, opened 2019-05-09
feat(core): gutenberg block support

bdbch:
This is a draft pull request for Gutenberg support as mentioned in #17. It's still WIP as I'm updating it. Help is appreciated!

### DIFF
--- a/lib/Api.php
+++ b/lib/Api.php
@@ -4,6 +4,7 @@ namespace Flynt;
 
 use ACFComposer\ACFComposer;
 use Flynt\ComponentManager;
+use Timber\Timber;
 use Dflydev\DotAccessData\Data;
 
 class Api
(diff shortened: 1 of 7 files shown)


lib/Init.php
```diff
@@ -18,6 +18,9 @@ public static function initTheme()
         // Set to true to load all assets from a CDN if there is one specified
         Asset::loadFromCdn(false);
 
+        // Set to true to use the new Gutenberg editor in combination with Flynt components
+        Defaults::useGutenberg(true);
```
bdbch: I wasn't sure if this should be handled in the Default class or API class. any thoughts? I also thought about having a Gutenberg class and use it to configure and define variables on how Gutenberg is configured in Flynt.

lib/Utils/TwigExtensionFlynt.php
```diff
@@ -25,11 +26,21 @@ public function getFunctions()
 
     public function renderFlexibleContent(Twig_Environment $env, $context, $fields, $withContext = true, $ignoreMissing = false, $sandboxed = false)
     {
-        $output = '';
-        foreach ((empty($fields) ? [] : $fields) as $field) {
-            $output .= $this->renderComponent($env, $context, ucfirst($field['acf_fc_layout']), $field, $withContext, $ignoreMissing, $sandboxed);
+        if (!Defaults::$useGutenberg) {
```
bdbch: Any thoughts on this one? As you can see I rely on `the_content` for Gutenberg.

Has some benefits f.e. RSS support, post excerpts, etc. but won't allow you to have multiple flexibleContent renderers on one page.

Any thoughts @domtra?

lib/Api.php
```diff
@@ -82,4 +84,63 @@ public static function registerHooks()
             );
         }, 10, 2);
     }
+
+    public static function registerBlocks($layouts)
+    {
+        // create block and field group for block
+        if (!function_exists('acf_register_block_type')) return null;
```
bdbch: I think this check could be done right before useGutenberg is defined. If the function does not exist it should just disable gutenberg again, will fix later.

bdbch:
Does anybody want to take a look?

cbirdsong:
@bdbch Has this stalled out?

bdbch:
@cbirdsong I could still pick it up but it's stale for now. You may give it a try if you want to!

cbirdsong:
@bdbch I assume even with this Flynt block data is stored in JSON blobs and not actual HTML, much like standard ACF blocks?

bdbch:
I used ACF Gutenberg blocks for this:
https://www.advancedcustomfields.com/blog/acf-5-8-introducing-acf-blocks-for-gutenberg/

They don't save the data as html but as a dataset as you'd expect.

steffenbew:
Decided not to support 2 approaches (standard ACF fields + ACF blocks) side by side. See #17 for further thoughts about why we don't support Gutenberg yet.

PR #33, opened 2019-06-21
NavigationSimple and NavigationMobile

Qakulukiam:


Components/NavigationMobile/Partials/Hamburger/index.twig
```diff
@@ -0,0 +1,4 @@
+<button class="hamburger" type="button" data-toggle-menu>
```
Qakulukiam: aria-label

Components/NavigationMobile/_style.scss
```diff
@@ -0,0 +1,40 @@
+$navigation-height: 75px;
+
+.pageWrapper {
+  padding-top: $navigation-height;
```
Qakulukiam: use height on ze component ja

PR #36, opened 2019-06-21
F slick/slider images centered

koukouchris:


Components/SliderImagesCentered/_style.scss
```diff
@@ -0,0 +1,156 @@
+[is='flynt-slider-images-centered'] {
+  padding: $component-spacing-mobile 0;
+
+  @media (min-width: $breakpoint-desktop) {
+    padding: $component-spacing-desktop 0;
+  }
+
+  .container {
+    @include center();
```
emcarru: you don't need the brackets if your're not passing args

PR #67, opened 2019-07-12
Block image text

nikola-bleech:
- Update component spacing
- Remove support for 'gif' files and add minimum image width
- Update image resize function syntax

### DIFF
--- a/Components/BlockImageText/_style.scss
+++ b/Components/BlockImageText/_style.scss
@@ -1,5 +1,13 @@
 [is='flynt-block-image-text'] {
-  margin: $component-spacing-desktop 0;
+  padding: $component-spacing-mobile 0;
+
+  @media (min-width: $breakpoint-tablet) {
+    padding: $component-spacing-tablet 0;
+  }
+
+  @media (min-width: $breakpoint-desktop) {
+    padding: $component-spacing-desktop 0;
+  }
 
   .container {
     @include center;
--- a/Components/BlockImageText/_style.scss
+++ b/Components/BlockImageText/_style.scss
@@ -9,10 +9,6 @@
     margin: $component-spacing-desktop 0;
   }
 
-  .container {
-    @include center;
-  }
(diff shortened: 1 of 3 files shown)


Components/BlockImageText/_style.scss
```diff
@@ -1,5 +1,13 @@
...
-  margin: $component-spacing-desktop 0;
+  margin: $component-spacing-mobile 0;
+
+  @media (min-width: $breakpoint-tablet) {
+    margin: $component-spacing-tablet 0;
+  }
+
+  @media (min-width: $breakpoint-desktop) {
+    margin: $component-spacing-desktop 0;
+  }
```
steffenbew: Please refer to #69.

Components/BlockImageText/functions.php
```diff
@@ -23,10 +23,11 @@
                 'name' => 'image',
                 'type' => 'image',
                 'preview_size' => 'medium',
-                'instructions' => '',
+                'instructions' => 'Minimum width of the image should be 1920px',
```
steffenbew: Why? I think this might _break_ it for many users and cases and would prefer no min size. If it's uploaded in smaller size, it should get scaled up in poor quality, but still work.

Components/BlockImageText/functions.php
```diff
@@ -23,10 +23,11 @@
...
                 'max_size' => 4,
+                'min_width' => 1920,
                 'required' => true,
-                'mime_types' => 'gif,jpg,jpeg,png'
+                'mime_types' => 'jpg,jpeg,png'
```
steffenbew: Why no gif anymore?

Components/BlockImageText/_style.scss
nikola-bleech: Ignored as per the following comment https://github.com/flyntwp/flynt/pull/69#discussion_r302961405

PR #69, opened 2019-07-12
Grid list steps

nikola-bleech:
Update code to new standard: Refactor the grid styles (use grid mixin)

### DIFF
--- a/Components/GridListSteps/_style.scss
+++ b/Components/GridListSteps/_style.scss
@@ -54,7 +54,7 @@
 
   .step {
     display: flex;
-    max-width: 400px;
+    max-width: 380px;
     padding-top: 10px;
   }
 
--- a/Components/GridListSteps/_style.scss
+++ b/Components/GridListSteps/_style.scss
@@ -1,5 +1,6 @@
 [is='flynt-grid-list-steps'] {
   $step-list-spacing: 20px;
+  $step-list-max-width: 380px;
 
   background-color: rgba(0, 0, 0, 0.03);
   padding: $component-spacing-mobile 0;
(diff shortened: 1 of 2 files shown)


Components/GridListSteps/_style.scss
```diff
@@ -1,6 +1,17 @@
...
-  padding: $component-spacing-desktop 0;
+  padding: $component-spacing-mobile 0;
+
+  @media (min-width: $breakpoint-tablet) {
+    padding: $component-spacing-tablet 0;
+  }
+
+  @media (min-width: $breakpoint-desktop) {
+    padding: $component-spacing-desktop 0;
+  }
```
steffenbew: Could you use the class `.centerMaxWidthContainer` on the `.container` element instead? (like used in the BlockWysiwygSidebar component)

Components/GridListSteps/_style.scss
```diff
@@ -15,29 +26,40 @@
...
+    @include reset-list;
 
-  .step {
-    max-width: 400px;
-    padding: 20px 10px 0;
+    align-items: center;
+    counter-reset: section;
+    flex-direction: column;
+    margin-left: auto;
+    margin-right: auto;
```
steffenbew: I assume this is about centering the grid items. If so, could we make use of the mixin `$alignment` option?

steffenbew:
In general: the changes add more CSS than previously, but the new mixins _should_ reduce complexity (besides harmonizing the approach). Could you please check if "it gets any simpler"?

Components/GridListSteps/_style.scss
nikola-bleech: Will use the class as suggested.
Please note that `.centerMaxWidthContainer` class does not contain`$component-spacing-mobile` variable

nikola-bleech: Class `.centerMaxWidthContainer` is dealing with the horizontal spacing. 
Styles marked here are defining the vertical spacing of the components.

I will add `.centerMaxWidthContainer` class in the markup.

Components/GridListSteps/_style.scss
nikola-bleech: `$alignment` option of the `grid` mixin deals only with `justify-content` property. 
Marked code, supplements the code generated by the mixin.

Components/GridListSteps/_style.scss
steffenbew: ah sure, understand.

nikola-bleech:
Revert the grid mixin to original styling.

PR #70, opened 2019-07-12
Grid content lists

nikola-bleech:
Add grid mixin and component spacing variables.


steffenbew:
Please refer to PR #69 for comments that might apply here as well.

nikola-bleech:
Reverted the grid mixin to the original styles.

PR #71, opened 2019-07-12
Added readme to ListIcons

petterastedt:


Components/ListIcons/README.md
```diff
@@ -0,0 +1,13 @@
+Title: List Icons
+
+---
+
+Category: list
```
nikola-bleech: The word `List` in the `Category` should be capitalized to make it consistent a cross flynt.

PR #74, opened 2019-07-14
feat(FieldVariables): added global field variables for theme options

steffenbew:


Components/BlockCookieNotice/index.twig
```diff
@@ -1,4 +1,4 @@
-<div is="flynt-block-cookie-notice" class="flyntComponent {{ layout }} {{ theme }}">
+<div is="flynt-block-cookie-notice" class="flyntComponent {{ layout }} {{ theme ? theme : 'themeDefault' }}">
```
domtra: short version: `{{ theme ?: 'themeDefault' }}`

domtra:
instead of defining the new class `Flynt\FieldVariables`, you could call `Flynt\Api::registerFields('FieldVariables', [...])`, and whenever you need them, call `Flynt\Api::loadFields('FieldVariables')`. This way, we would reuse existing functionality.
If you think that is confusing, because `FieldVariables` is not a component (and we usually use `registerFields`) for components, please put the file `fieldVariables.php` into `lib` instead of `inc` and rename it to `FieldVariables.php` (capital F). Since the file contains only a class, it should be auto loaded when needed and not for every request.

PR #78, opened 2019-07-16
List search results

nikola-bleech:


Components/ListSearchResults/Partials/Header/_style.scss
```diff
@@ -1,59 +1,87 @@
...
+
+      @media (min-width: $breakpoint-mobile-horizontal) {
+        margin-bottom: 0.8em;
+      }
 
-    padding: $container-padding-desktop 0;
+      @media (min-width: $breakpoint-tablet) {
+        margin-bottom: 1.75em;
+      }
+    }
```
steffenbew: This doesn't look like what I expected. Isn't it just a wysiwyg, so you could just add an h1 or h2 and would look like intended, without any additional styling (like it was before)?

Components/ListSearchResults/Partials/Pagination/_style.scss
```diff
@@ -36,4 +12,12 @@
       }
     }
   }
+
+  &-link {
+    &[disabled] {
+      cursor: default;
+      opacity: 0.65;
+      pointer-events: none;
```
steffenbew: I would prefer not showing buttons when they're not needed. You will probably need the floating approach then again.

Components/ListSearchResults/Partials/Results/_style.scss
```diff
@@ -1,14 +1,20 @@
 .searchList {
+  background-color: #fff;
```
steffenbew: Please use the `$color-background` variable from `_variables.scss` instead.

Components/ListSearchResults/Partials/Results/_style.scss
```diff
@@ -49,8 +55,16 @@
...
     margin-bottom: 0;
     padding-bottom: 0;
   }
 }
+
+.postTitle {
+  font-size: 1rem;
+}
+
+.postLink {
```
steffenbew: These 2 classes don't look like naming conventions according to maintainable css to me. Please refer to https://maintainablecss.com/chapters/modules/.

Components/ListSearchResults/_style.scss
```diff
@@ -1,13 +1,51 @@
...
+    @media (min-width: $breakpoint-desktop) {
+      max-width: $content-max-width + $container-padding-desktop *2;
+    }
+  }
+
+  &.themeDark {
+    .header {
+      &-title {
+        > * {
+          color: #fff;
```
steffenbew: I believe we shouldn't need this when just using standard headline elements.

Components/ListSearchResults/_style.scss
```diff
@@ -1,13 +1,51 @@
...
     @import 'Partials/Results/style';
+
+    max-width: $content-max-width + $container-padding-mobile *2;
+
+    @media (min-width: $breakpoint-tablet) {
+      max-width: $content-max-width + $container-padding-tablet *2;
+    }
+
+    @media (min-width: $breakpoint-desktop) {
+      max-width: $content-max-width + $container-padding-desktop *2;
```
steffenbew: I think it might be easier to just have another wrapping container and give it the class `centerContentMaxWidth` in HTML, so you will be able to remove all this css.

Components/ListSearchResults/_style.scss
```diff
@@ -1,13 +1,51 @@
...
+  &.themeHero,
+  &.themeLight {
+    padding: $component-spacing-mobile 0;
+
+    @media (min-width: $breakpoint-tablet) {
+      padding: $component-spacing-tablet 0;
+    }
+
+    @media (min-width: $breakpoint-desktop) {
+      padding: $component-spacing-desktop 0;
```
steffenbew: Nice! Still need to add this on a global level.

steffenbew:
When there's no excerpt, the "Read More" link gets pulled into the first line and is bigger. I think it should still be below and smaller.
Bildschirmfoto 2019-07-16 um 20 27 05

Components/ListSearchResults/Partials/Header/_style.scss
nikola-bleech: Was no option to change to heading from the admin panel on my local machine. 
Will remove this one.
srch

steffenbew: Please click the switch to "Visual" on the top right and you'll have the options.

Components/ListSearchResults/_style.scss
nikola-bleech: Will leave this for the global task.

nikola-bleech:
Implemented the suggested changes.

Components/ListSearchResults/Partials/Header/_style.scss
```diff
@@ -1,59 +1,83 @@
...
   &-input {
-    background-color: #fff;
-    border: 2px solid $color-accent;
+    border: 0;
     color: #9b9b9b;
-    font-size: 30px;
-    font-weight: 700;
-    height: 80px;
+    flex-shrink: 1;
+    font-size: 0.875rem;
```
steffenbew: I feel the font size is very small now. Would like to deviate from the design and keep the default size.

Components/ListSearchResults/Partials/Header/_style.scss
```diff
@@ -1,59 +1,83 @@
...
+    padding-left: 50px;
+    padding-right: $search-form-spacing;
     position: relative;
   }
 
   &-input {
-    background-color: #fff;
-    border: 2px solid $color-accent;
+    border: 0;
     color: #9b9b9b;
```
steffenbew: Please remove the color that was set previously. Should be the default color.

Components/ListSearchResults/Partials/Results/_style.scss
```diff
@@ -1,14 +1,20 @@
 .searchList {
+  background-color: $color-background;
+  border-radius: 4px;
+  box-shadow: 0 1px 26px 0 rgba(0, 0, 0, 0.06);
   list-style: none;
+  margin-top: 1.25rem;
```
steffenbew: I think this should be 20px instead, to make it consistent with the padding of the pagination used later.

Components/ListSearchResults/Partials/Pagination/_style.scss
```diff
@@ -1,29 +1,38 @@
...
 
-  div {
-    display: block;
-    margin: 10px 2px;
+  // Float clearfix
+  &::after {
+    clear: both;
+    content: '';
+    display: table;
+  }
```
steffenbew: I like a shorter fix: Instead of adding an ::after element, just add `overflow: auto;` to the .pagination element.

Components/ListSearchResults/Partials/Pagination/_style.scss
```diff
@@ -1,29 +1,38 @@
...
+
+  &-nav {
+    width: 100%;
```
steffenbew: I would rather stick to regular button widths and always float the buttons. This will also simply the css on this one.

Components/ListSearchResults/Partials/Results/_style.scss
```diff
@@ -48,9 +54,33 @@
     }
   }
 
-  .excerpt {
-    color: #9b9b9b;
+}
+
+.post {
+  &-title {
```
steffenbew: I feel this should rather be a default heading selector with default styles (like a `h3` tag). We might need to review the overall font-sizes of this component then, but I think it's better to keep it generic than making a lot of custom typographic changes just for this component.

Components/ListSearchResults/Partials/Results/index.twig
```diff
@@ -6,14 +6,14 @@
...
       {% endif %}
-      <div class="content">
-        <a class="postLink" href="{{ post.link }}">
-          <h3 class="postTitle">{{ post.title }}</h3>
-        </a>
+      <div class="post">
+        <a class="post-title" href="{{ post.link }}">{{ post.title }}</a>
         {% if post.preview|length > 10 %}
-          <p class="excerpt">{{ post.preview.length(40) }}</p>
+          <p class="post-excerpt">{{ post.preview.length(40) }}</p>
```
steffenbew: You could extend it to `{{ post.preview.length(40).read_more(false) }}` to not have a read more link and then add the read more link manually via `<a href="{{ post.link }}">{{ readMore }}</a>`. That way, I think you won't need the extra post-footer element.

Components/ListSearchResults/Partials/Header/_style.scss
```diff
@@ -1,59 +1,83 @@
+$search-form-spacing: 6px;
+
 .header {
   &-title {
```
steffenbew: Could you please remove the -title element and instead implement a preContent approach like used in other recently developed components (e.g. BlockCountUp)?

Components/ListSearchResults/Partials/Header/_style.scss
```diff
@@ -1,59 +1,83 @@
...
+    position: relative;
+    width: 100%;
+
+    @media (min-width: $breakpoint-mobile-horizontal) {
+      flex-direction: row;
+    }
+  }
+
+  &-box {
+    flex-grow: 1;
```
steffenbew: How about instead of adding padding to the box, adding padding to the input instead? That way, I could imagine we could save some code and wouldn't need the autofill hack as well.

Components/ListSearchResults/Partials/Pagination/_style.scss
nikola-bleech: There is a high probability that on `320px` screen width, buttons won't have enough space to float as suggested, if different content is used (screenshot).

no-space

Components/ListSearchResults/Partials/Header/_style.scss
nikola-bleech: Input does not actually fill in the entire "box" area (please see screesnshot). It was necessary to use this approach, since the width of the button might vary with different translations/content.

When the `autofill` styling kicks in, the `input` element gets a different color to the one of the box and looks a bit weird, in my opinion.

search

Components/ListSearchResults/Partials/Pagination/_style.scss
steffenbew: We might use the smaller button style instead.

Components/ListSearchResults/Partials/Header/_style.scss
steffenbew: Understand with the flex-grow. I would probably use `width: 100%` instead, but doesn't make a difference then. Just double checking: what's the `flex-shrink: 1` on the input doing?

In terms of the autofill, I would like to keep the browser functionality, as that's what people expect and are used to. There might be exceptions to the rule, but I feel making the spacing adjustments as suggested, we can still make it look nice and keep the browser default.

nikola-bleech: I forgot to mention that, I've already removed the `flex-shrink: 1` from the `input`.

Will remove the autofill code and also implement the padding as suggested.

PR #80, opened 2019-07-17
refactor(List: Social): Add helper class.

nikola-bleech:
Use global helper class to center the container.


nikola-bleech:
Hi Steffen,
I've cleaned up code as you suggested.

PR #86, opened 2019-07-18
Base Style & TinyMCE Configuration

steffenbew:
- changed link color to white for hero theme
- Introduced paddingBox helper class and removed `overflow: auto` from center mixin
- Added arrow styled link option to WYSIWYG
- reduced tinyMce toolbar options to default and basic + removed toolbar spec from function.php files


assets/styles/_base.scss
```diff
@@ -107,12 +107,17 @@ body {
...
+  a,
+  .feather {
     color: $color-background;
   }
 
-  .feather {
-    color: $color-background;
+  @media (min-width: $breakpoint-tablet) {
+    a:hover,
+    a:focus {
```
nikola-bleech: Perhaps nesting the states inside a single `<a>` tag selector would be better. 
Like so:
```
a {
    &:hover,
    &:focus {
      ...
    }
}
```

templates/twig/basestyle.twig
```diff
@@ -24,10 +24,11 @@
         <p>Drops of rain could be heard hitting the pane, which made him feel quite sad. "How about if I sleep a little bit longer and forget all this <strong>nonsense</strong>", he thought, but that was something he was unable to do because he was used to sleeping on his right, and in his present state couldn't get into that position. However hard he threw himself onto his right, he always rolled back to where he was.</p>
         <p>He must have tried it a hundred times, shut his eyes so that he wouldn't have to look at the floundering legs, and only stopped when he began to feel a mild, dull pain there that he had never felt before.</p>
         <hr>
-        <div>
+        <div class="themeDefault centerMaxWidthContainer">
           <h2>Default Theme</h2>
-          <p>Not using theme's classes allows you to style a group of content with a default background and default text color.</p>
+          <p>Using the themeDefault class allows you to style a group of content with a default background and default text color.</p>
            <p>
+            <a href="#">Regular link</a><br><br>
```
nikola-bleech: When using an anchor tag that navigates nowhere, perhaps it would be better instead of `#`, to use one of the following in the `href`:
- `#!` - reference to an id with a name of `!`
- `javascript:void(0)`

in order to prevent bouncing to the top of the page once the link has been clicked.

inc/tinyMce.php
```diff
@@ -84,15 +84,35 @@ function getConfig()
                 'title' => 'Buttons',
                 'icon' => '',
                 'items' => [
+                    [
+                        'title' => 'Button',
+                        'classes' => 'button',
+                        'selector' => 'a,button'
```
nikola-bleech: Perhaps the selectors should have a white space after the comma? -> `a, button`

assets/styles/_base.scss
steffenbew: I didn't like the nesting and decided against it because I don't expect needing any styles on the a tag and thus I feel it wouldn't add value.

templates/twig/basestyle.twig
steffenbew: Good point, will change.

inc/tinyMce.php
steffenbew: Good point, though I followed the TinyMCE way without separator and thus would keep it like that: https://www.tiny.cloud/docs/configure/content-formatting/#example

PR #87, opened 2019-07-18
Grid posts teaser

nikola-bleech:
- register new component
- add new component layout


Components/GridPostsTeaser/_style.scss
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @media (min-width: $breakpoint-desktop) {
+    padding: $component-spacing-desktop 0;
+  }
+
+  .preContent {
+    @include reset-content-margins;
+
+    margin-bottom: 4.375rem;
```
steffenbew: Looking at other components, the typical margin-bottom is 1.5rem or 50px. In GridPostsLatest we're using then 3rem, which is also weird (think that was me). We should unify this in general. So what can we agree on. 50px?

Components/GridPostsTeaser/_style.scss
```diff
@@ -0,0 +1,148 @@
...
+    text-align: center;
+  }
+
+  .postContent {
+    margin-top: $container-padding-mobile + 10;
```
steffenbew: I would vote to make this the same as the preContent margin-bottom and also align with having breakpoints (or rather: not having breakpoints).

Components/GridPostsTeaser/_style.scss
```diff
@@ -0,0 +1,148 @@
...
+    text-decoration: none;
+    width: 100%; // IE11 bug fix
+
+    @media (min-width: $breakpoint-mobile-horizontal) {
+      transition: box-shadow 0.2s ease-in;
+    }
+
+    &:hover {
+      @media (min-width: $breakpoint-mobile-horizontal) {
+        box-shadow: none;
```
steffenbew: Nice idea! Personally, I don't like it though. I feel it should be rather the other way around, that hovering over it, the shadow would get stronger and the item bigger. But right now, I would rather stick with no hover (as I don't have a better idea).

Components/GridPostsTeaser/_style.scss
```diff
@@ -0,0 +1,148 @@
...
+      @include reset-list;
+
+      align-items: center;
+      display: flex;
+      font-size: 0.8125rem;
+      justify-content: space-between;
+      margin-bottom: 0.625rem;
+    }
+
+    &-pill {
```
steffenbew: Could / should this be a global base style class?

Components/GridPostsTeaser/functions.php
```diff
@@ -0,0 +1,137 @@
...
+        'placement' => 'top',
+        'endpoint' => false
+    ],
+    [
+        'label' => '',
+        'name' => 'labels',
+        'type' => 'group',
+        'sub_fields' => [
+            [
+                'label' => 'Reading Time Label',
```
steffenbew: This field seems to be not used in this component, so should be removed.

Components/GridPostsTeaser/functions.php
```diff
@@ -0,0 +1,137 @@
...
+            ],
+            [
+                'label' => 'All Posts Label',
+                'name' => 'allPosts',
+                'type' => 'text',
+                'default_value' => 'See More Posts',
+                'required' => 1,
+            ],
+            [
+                'label' => 'Read More Label',
```
steffenbew: This field seems to be not used in this component, so should be removed.

Components/GridPostsTeaser/index.twig
```diff
@@ -0,0 +1,49 @@
...
+                {{ item.thumbnail|resize(1280, 853) }} 1280w,
+                {{ item.thumbnail|resize(1050, 700) }} 1050w,
+                {{ item.thumbnail|resize(750, 500) }} 750w,
+                {{ item.thumbnail|resize(600, 400) }} 600w,
+                {{ item.thumbnail|resize(480, 320) }} 480w,
+                {{ item.thumbnail|resize(312, 208) }} 312w"
+              data-sizes="auto"
+              alt="{{ item.thumbnail.alt|e }}">
+            <div class="content-inner">
+              <ul class="content-header">
```
steffenbew: A list doesn't feel right to me here. Maybe a `<header>`?

Components/GridPostsTeaser/index.twig
```diff
@@ -0,0 +1,49 @@
...
+              <li class="content-block">
+              <span class="content-pill">{{ item.category.name }}</span>
+              </li>
+              <li>
+              <dl class="content-date">
+              <dt>
+                  <i data-feather="clock"></i>
+              </dt>
+              <dd>{{ item.date }}</dd>
```
steffenbew: I think all the above should rather be in a `<time>` tag.

steffenbew:
@nikola-bleech It would be great if you could check the other post components (like GridPostsSlider, GridPostsLatest) in terms of consistency, to adapt GridPostsTeaser and to figure out where we will need to make changes in other components still.

Components/GridPostsTeaser/_style.scss
nikola-bleech: The design uses mostly `70px` for this kind of spacing. If we are to follow the designer's recommendation I would then suggest using either the exact value, which is `4.375rem (70px)`.

The real question is: is there a particular reason to drop the original `70px` and go for `50px`?

70px

Components/GridPostsTeaser/_style.scss
nikola-bleech: This should definitely be a global class.
In fact, in my opinion, all the elements that are commonly used a cross the web site should be defined in global scope.
The logic is, if we can use a `button` class in global scope, what is the reason not to use other elements?
For example, card-looking element is being re-defined in every component even though all the instances of it share quite a bit of common properties, like the `padding` and `background-color` (please see screenshot).
---
I would not use the base style file for this, but rather block-specific files: `buttons.scss, pills.scss, etc.`

glob

PR #91, opened 2019-07-22
refactor(List: Facts): Remove reduntant component

nikola-bleech:
List: Facts component is a duplicate of the Block: Count Up component.


steffenbew:
@nikola-bleech Did you check if ListFacts needs to be removed from the fieldGroups as well?

nikola-bleech:
@steffenbew 
Missed de-registering it there. Thank you for pointing that out.

PR #93, opened 2019-07-24
GlobalContentClassesWeek30

mrnschrmm:


assets/styles/_base.scss
```diff
@@ -716,3 +716,39 @@ select {
     }
   }
 }
+
+/**
+* 4.0 Content
+*/
+
+.flyntComponent {
```
steffenbew: Please move the whole thing to the upper section _2.0 Layout_ and remove the wrapping .flyntComponent class. I understand that you would like to regroup elements, but lets do that in one go for all existing and upcoming helper classes at a later point in time.

assets/styles/_base.scss
```diff
@@ -716,3 +716,39 @@ select {
...
+    @media (min-width: $breakpoint-tablet) {
+      margin-bottom: $container-padding-tablet;
+    }
+
+    @media (min-width: $breakpoint-desktop) {
+      margin-bottom: $container-padding-desktop;
+    }
+  }
+
+  .postContent {
```
steffenbew: I think we should consider adding `@include reset-content-margins;` here as well. What do you think?

steffenbew:
@mrnschrmm 

1. I believe we should remove the `centerContentMaxWidth` classes from `preContent` elements, as the behavior seems to be added twice otherwise. 
2. Could you please review and unify the markup of preContent and postContent elements (sometimes it's on one line, sometimes on multiple lines, different indentation, etc.)?
3. While looking into that, I noticed `ListIcons` is using its own markup for preContent, so should be adjusted in markup and css.

PR #98, opened 2019-07-25
Grid posts slider

nikola-bleech:
I eventually decided to make the slider full width of the screen, as the `Slider: Images Centered` is using the same approach. I've added the padding on either side of the slider which is equal to the spacing between the slides them selves.

Changed the breakpoint of the slider to give more "breathing" room to the cards on smaller screens.
ssl-ider


Components/GridPostsSlider/script.js
```diff
@@ -36,21 +36,22 @@ class GridPostsSlider extends window.HTMLDivElement {
...
-        },
-        768: {
           slidesPerView: 2
         },
-        1024: {
+        768: {
           slidesPerView: 3
+        },
+        1280: {
+          slidesPerView: 4
```
steffenbew: should we add another one for super large screens? 6 items from 1440?

nikola-bleech: That is a good point. I've decided to go with 5 cards at 1440px screen width, instead of 6, because the they look a bit cluttered at that particular break 
ssliider
point. I did, however, add a layout with 6 cards at 1920px screen width.

PR #99, opened 2019-07-25
Grid posts teaser

nikola-bleech:
Include changes not pushed in the previous PR (https://github.com/flyntwp/flynt/pull/87)

---

@steffenbew 
I've left the `pill` element in the cards, as in my opinion it gives a quick and easy overview on the type of the post in question. The fact that in some cases the text has to be trimmed with trailing dots `...`  is not a huge deal (in my humble opinion), as the category name needs to be quite long (over 17 characters in length).

I have also changed the color of the `date` element from light grey to dark-blue, as I think it's easier to spot. I decided against the orange color since we use that one on the links, and the date element is not a link by it self.


Components/GridPostsTeaser/_style.scss
```diff
@@ -85,63 +67,40 @@
...
-        }
+  .time {
+    align-items: center;
+    display: flex;
+    margin: 0;
+
+    &-icon,
+    &-main {
+      color: $color-background-dark;
+      margin: 0;
```
nikola-bleech: remove

Components/GridPostsTeaser/_style.scss
```diff
@@ -85,63 +67,40 @@
...
-  &.themeDefault {
-    .content {
-      &:hover {
-        @media (min-width: $breakpoint-mobile-horizontal) {
-          box-shadow: 0 1px 8px 0 rgba(0, 0, 0, 0.06);
-        }
+  .time {
+    align-items: center;
+    display: flex;
+    margin: 0;
```
nikola-bleech: remove?

Components/GridPostsTeaser/_style.scss
```diff
@@ -13,25 +13,17 @@
...
-
-    @media (min-width: $breakpoint-desktop) {
-      margin-top: $container-padding-desktop + 10;
-    }
   }
 
   .grid {
     @include reset-list;
     @include grid(1, center);
+    margin-bottom: 4.375rem;
```
nikola-bleech: add new line

PR #112, opened 2019-07-29
LayoutSinglePostWeek31

mrnschrmm:


Components/BlockPostHeader/Partials/Image/_style.scss
```diff
@@ -0,0 +1,12 @@
+.image {
+  display: block;
+  height: auto;
+  margin: 0;
+  width: 100%;
```
steffenbew: I think you don't need all of the above. Instead though, please use the same markup and styling as in `BlockImage` (with <figure>).

Components/BlockPostHeader/Partials/Image/index.twig
```diff
@@ -0,0 +1,13 @@
+<div class="image">
+  <img class="image-tag lazyload"
+    src="{{ image.src|resizeDynamic(840) }}"
+    srcset="{{ placeholderImage(840, 840 / image.aspect) }}"
+    data-srcset="
+      {{ image.src|resizeDynamic(840) }} 840w,
```
steffenbew: Please prepare also high resolution sizes.

Components/BlockPostHeader/_style.scss
```diff
@@ -0,0 +1,109 @@
+[is='flynt-block-post-header'] {
+
+  // ##BLOCK-POST-HEADER
+  // =============================================================
+
+  // SPACING
+  // -------------------------------------------------------------
+
+  .container {
+    @include center;
```
steffenbew: Instead, please use the helper class in the markup directly.

Components/BlockPostHeader/_style.scss
```diff
@@ -0,0 +1,109 @@
...
+      margin-right: 0.25rem;
+    }
+
+    &-readingtime {
+      margin-left: 0.25rem;
+    }
+
+    &-intro {
+      p {
+        font-size: 1.125rem;
```
steffenbew: I would prefer using one of our headline classes instead, like at the top in the BaseStyle.

Components/BlockPostHeader/index.twig
```diff
@@ -0,0 +1,25 @@
+<div class="flyntComponent" is="flynt-block-post-header">
+  <div class="container">
+    <div class="blockPostHeader">
+      <div class="blockPostHeader-title">
+        <h1 class="blockPostHeader-h1">{{ post.title }}</h1>
+      </div>
+      <div class="blockPostHeader-meta">
+        {% if post.date %}
+          <time class="blockPostHeader-date">{{ post.date|date }}</time> {{ generalTranslatable.authorDivider|e }}
```
steffenbew: I think it's missing the timestamp here.

Components/BlockPostHeader/index.twig
```diff
@@ -0,0 +1,25 @@
...
+        {% endif %}
+        {{ generalTranslatable.postedByLabel|e }} <a class="blockPostHeader-author" href="{{ post.author.link|e('esc_url') }}">{{ post.author.name|capitalize }}</a>
```
steffenbew: Not sure we should capitalize. I feel it should be the actual nick name.

Components/BlockPostHeader/index.twig
```diff
@@ -0,0 +1,25 @@
...
+        {{ generalTranslatable.postedInLabel|e }} <a class="blockPostHeader-category" href="{{ post.category.link|e('esc_url') }}">{{ post.category }}</a>
+        {{ generalTranslatable.readingtimeDivider|e }} <span class="blockPostHeader-readingtime">{{ post.content|readingtime }} {{ generalTranslatable.readingtimeLabel|e }}</span>
+      </div>
+      <div class="blockPostHeader-intro">
+        <p>{{ post.preview.read_more(false) }}</p>
```
steffenbew: Does this work now also with not having a manual excerpt, or does it show the beginning of the actual post content instead?

Components/BlockPostHeader/index.twig
```diff
@@ -0,0 +1,25 @@
...
+      </div>
+      {% if post.thumbnail %}
+        <div class="blockPostHeader-media">
+          {% include 'Partials/Image/index.twig' with {image: post.thumbnail} only %}
```
steffenbew: Couldn't we also render the image component here and pass the image as an option as well?

assets/styles/_base.scss
```diff
@@ -135,6 +135,45 @@ body {
...
+
+      @media (min-width: $breakpoint-tablet) {
+        margin-top: 3rem;
+      }
+    }
+
+    &-main,
+    &-related {
+      > .flyntComponent {
+        margin: 0;
```
steffenbew: This shouldn't be the case. At most the first and last components margin-/padding-top / -bottom.

Components/BlockPostFooter/_style.scss
```diff
@@ -0,0 +1,170 @@
...
+      &:not(:last-child) {
+        margin: 0 8px 0 0;
+
+        @media (min-width: $breakpoint-tablet) {
+          margin-right: 12px;
+        }
+      }
+    }
+
+    &-tagLink {
```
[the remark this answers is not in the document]
mrnschrmm: I left what is still necessary to let it look like before or like in GridPostArchive. I assume the pill class isn't complete yet.

Components/BlockPostHeader/Partials/Image/_style.scss
mrnschrmm: Integrated BlockImage ... but generally there is no CSS at all related to the image or structures shown in the lazysizes docs.

PR #115, opened 2019-08-01
refactor(List: Search Results): Review the layout

nikola-bleech:
- Use 'boxPadding' helper class. 
- Remove the image from the results. 
- Revert the search results font size to default. 
- Align the search icon with results copy.
- Add vertical separation between `Read more` link and the result text.


nikola-bleech:
- Remove search icon
- Update screenshot

Components/ListSearchResults/Partials/Header/_style.scss
```diff
@@ -50,14 +49,11 @@ $search-form-spacing: 6px;
     margin: 0;
     min-height: 44px;
     outline: none;
-    padding: 0 1.25rem 0 3.125rem;
-    transition: all 0.1s cubic-bezier(0.175, 0.885, 0.32, 0.275);
+    padding: 0 1.25rem 0 (20px - 6); // .boxPadding (mobile) - $search-form-spacing
```
steffenbew: I feel all margins should be in pixels then, mixed is confusing to me. Also, why don't you just use the actual variables instead of commenting them?

Components/ListSearchResults/Partials/Header/index.twig
```diff
@@ -6,7 +6,6 @@
     <div class="searchForm-field boxShadow">
       <div class="searchForm-box">
         <label class="searchForm-label" for="search-input">
```
steffenbew: I think this (and the corresponding) CSS should be removed.

Components/ListSearchResults/Partials/Header/_style.scss
nikola-bleech: Padding was originally meant to be expressed in `rem`, but since the side spacing needs to follow the `.boxPadding` values which are expressed in `px`. 
I will refactor the padding to `px` and introduce a variable that takes care of the box spacing.

steffenbew: Ah, I thought we had a variable for `.boxPadding` already. I don't think we need to introduce one then. Would be fine with converting everything to px instead and just using the `$search-form-spacing` variable that's available already.

Components/ListSearchResults/Partials/Header/index.twig
nikola-bleech: I would advise against this, as it has negative impact on accessibility - screen readers in particular, as described in [this article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#Why_you_should_use_labels).

Would be great to hear if you agree with this approach.

steffenbew: I'm totally in for labels, but in order to be accessible, they need to be visible and have text. I think the article doesn't say anything different. So to my understanding an empty invisible label doesn't add anything and should be removed. An arial-label could be introduced instead onto the input field, but I don't have enough knowledge and experience about that to make a positive decision.

nikola-bleech:
- remove label tag
- update styles with variables

steffenbew:
@nikola-bleech please see my comments on https://bleech.monday.com/boards/268294353/pulses/282806492

nikola-bleech:
- Cleaned up the markup (removed unused code)
- Made "read me" link right aligned - looks easier to spot now
- Updated the "README" file to reflect current changes

PR #116, opened 2019-08-01
Grid posts latest

nikola-bleech:
Refactor the layout so that the text fills the entire height of the card.

__before
---
__after

### DIFF
--- a/Components/GridPostsLatest/_style.scss
+++ b/Components/GridPostsLatest/_style.scss
@@ -72,9 +72,9 @@
 
       .feather {
         color: #7b838d;
-        height: auto;
+        height: 1rem;
         margin-right: 2px;
-        width: 16px;
+        width: 1rem;
       }
     }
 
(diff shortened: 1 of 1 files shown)


Components/GridPostsLatest/_style.scss
```diff
@@ -60,9 +63,15 @@
...
       @include reset-content-margins;
+
+      flex-grow: 1;
     }
 
     &-link {
+      display: flex;
+      flex-direction: column;
+      flex-grow: 1;
+      justify-content: space-between;
```
steffenbew: Do we need this line, as we don't want to have space anyways?

Components/GridPostsLatest/_style.scss
```diff
@@ -43,11 +43,14 @@
 
   .content {
     display: flex;
+    flex-direction: column;
     flex-wrap: wrap;
     height: 100%;
+    justify-content: space-between;
```
steffenbew: see comment below

Components/GridPostsLatest/_style.scss
nikola-bleech: You are right, the line is un-necessary and will be removed before merging.

PR #118, opened 2019-08-02
Hero image text

nikola-bleech:
- reduce content padding on mobile (has more space to expand to)
- remove the retina image size for desktop devices
- re-define mobile image sizes and the breakpoint
- update component screenshot (old one had an image that was only 1000px in width and looked blurry on larger screens)

---
New `source` breakpoint value was derived from the container width (1440px /2 = 720px). This allows to have the retina display support on mobile, while still loading the appropriate image sizes.


steffenbew:
@nikola-bleech Thanks for spotting the mobile padding! I understand your thoughts about the new sizes, however I would be happy to have a desktop retina image as well. And for the new sizes: I feel a general issue is that as soon as you create a different crop size (same height, but different width), the format of the image content visually changes. It's hard to prevent it while optimizing for loading time, but we should keep the different sizes to a minimum (that's why on mobile I only had 1 + 1 retina, but I think I didn't consider it for desktop). Not sure which yet, I have to think about it again as well and will take over.

PR #120, opened 2019-08-05
GridPostsArchiveUpdate01Week32

mrnschrmm:


Components/GridPostsArchive/index.twig
```diff
@@ -12,7 +12,7 @@
     <ul class="posts">
       {% for post in posts %}
         <li class="post">
-          <a class="post-link" href="{{ post.link }}"{% if post.target == '_blank' %} target="_blank" rel="noreferrer noopener"{% endif %} title="{{ post.title|e }}">
+          <a class="post-link" href="{{ post.link }}" {{ post.link.target == '_blank' ? 'target="_blank" rel="nofollow noreferrer noopener"' : 'rel="nofollow"' }} title="{{ post.title }}">
```
steffenbew: Why did you add `nofollow`? I think if I would propose a change, that would be removing the `post.link.target` check and `target="_blank"` stuff entirely, as I think there's no option to set that with this component. It will (and should) always be internal links.

PR #132, opened 2019-10-22
fix(GoogleAnalyticsOptIn): Removed block cookie notice from doucment …

MyEmpireOfDuctTape:
…and added googleanalyticsoptin instead


Components/GoogleAnalyticsOptIn/script.js
```diff
@@ -0,0 +1,152 @@
...
+    this.acceptTracking = this.acceptTracking.bind(this)
+    this.declineTracking = this.declineTracking.bind(this)
+    this.showModal = this.showModal.bind(this)
+    this.hideModal = this.hideModal.bind(this)
+  }
+
+  bindEvents () {
+    this.$acceptButton.on('click', this.acceptTracking)
+    this.$declineButton.on('click', this.declineTracking)
+    this.$optOutLink.on('click', this.showModal)
```
usingthesystem: how about adding the click event listener like that?
$(body).on('click', '.googleOptOutLink',this.showModal)
That way it would also work when the link is somehow dynamically added to the dom at any point in time

Components/GoogleAnalyticsOptIn/script.js
```diff
@@ -0,0 +1,152 @@
...
+    const scriptUrl = `https://www.googletagmanager.com/gtag/js?id=${this.gaId}`
+    this.GTAGscriptElem = document.createElement('script')
+    this.GTAGscriptElem.async = true
+    this.GTAGscriptElem.type = 'text/javascript'
+    this.GTAGscriptElem.src = scriptUrl
+    document.head.appendChild(this.GTAGscriptElem)
+  }
+
+  addGTAGFunction () {
+    const scriptContent = `window.dataLayer = window.dataLayer || [];
```
usingthesystem: could we just execute the function instead of adding it as a script tag in the head?

Components/GoogleAnalyticsOptIn/script.js
```diff
@@ -0,0 +1,152 @@
...
+    gtag('js', new Date());
+    gtag('config', '${this.gaId}', { 'anonymize_ip': true });`
+    this.GTAGscriptElemFunction = document.createElement('script')
+    this.GTAGscriptElemFunction.type = 'text/javascript'
+    this.GTAGscriptElemFunction.text = scriptContent
+    document.head.appendChild(this.GTAGscriptElemFunction)
+  }
+
+  addGTAGFunctionFallback () {
+    const scriptContent = `function gtag() {
```
usingthesystem: could we just execute the function instead of adding it as a script tag in the head?

Components/GoogleAnalyticsOptIn/script.js
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  declineTracking () {
+    Cookies.set(this.cookieName, false, { expires: Number(this.props.expiryDays), path: window.location.pathname })
+    this.hideModal()
+    if (this.gaId) {
+      Cookies.set(this.disableStr, true)
+      window[this.disableStr] = true
+      if (typeof this.GTAGscriptElem !== 'undefined') {
+        this.GTAGscriptElem.remove()
```
usingthesystem: GTAGscriptElem i would right the variable name like that
GtagScriptElement 
i think it is easier to read and in general we try to avoid abbreviations

package.json
```diff
@@ -114,7 +114,9 @@
       "postbump": "gulp replaceVersion && git add ."
     }
   },
-  "dependencies": {},
+  "dependencies": {
+    "js-cookie": "^2.2.1"
```
usingthesystem: plz add the js-cookie dependency as devDependency

Components/GoogleAnalyticsOptIn/script.js
MyEmpireOfDuctTape: I've added this and removed the $optoutlink propery.

Components/GoogleAnalyticsOptIn/script.js
MyEmpireOfDuctTape: We can with a  few modifications

Components/GoogleAnalyticsOptIn/script.js
MyEmpireOfDuctTape: Same, with a few modifications

package.json
MyEmpireOfDuctTape: Done

domtra:
@MyEmpireOfDuctTape can this be closed?

PR #145, opened 2019-10-30
fix(ListIcons): output url when no title text

emcarru:


Components/ListIcons/index.twig
```diff
@@ -15,7 +15,9 @@
         </div>
         {% endif %}
         <div class="item-linkWrapper">
-          <a href="{{ item.link.url }}" class="item-link button--link">{{ item.link.title }}</a>
+          <a href="{{ item.link.url }}" class="item-link button--link">
```
steffenbew: While on it: could you please add the target?

steffenbew: Oh, and please add `.button` to the classes, as I feel we shouldn't use a modifier without it's base class. (Yes, I'm aware this will change the appearance.)

emcarru: both done now!

PR #147, opened 2019-10-31
feat(lib/FieldVariables): move to class in lib

emcarru:
- Prevents ordering issues when using inc files inside fieldVariables which we've had a few times
- Removes another use case for loadFields, makes it simpler
- Personal preference: it's easier to read :-)


domtra:
closed due to https://github.com/flyntwp/flynt/pull/179

PR #161, opened 2019-11-07
Replace feather icons select fields with simple text field

emcarru:


inc/fieldVariables.php
```diff
@@ -21,5 +21,12 @@
             'themeDark' => 'Dark',
             'themeHero' => 'Hero'
         ]
-    ]
+    ],
+    'icon' => [
+        'label' => 'Icon',
+        'name' => 'icon',
+        'type' => 'text',
+        'instructions' => 'Enter a valid icon name from Feather Icons (e.g. `check-circle`). Visit https://feathericons.com/ to see the all icons with their corresponding names.',
```
steffenbew: How about linking "Feather Icons" to the website and remove the second sentence to make it short?

emcarru: Sure, updated.

inc/fieldVariables.php
```diff
@@ -26,7 +26,7 @@
         'label' => 'Icon',
         'name' => 'icon',
         'type' => 'text',
-        'instructions' => 'Enter a valid icon name from Feather Icons (e.g. `check-circle`). Visit https://feathericons.com/ to see the all icons with their corresponding names.',
+        'instructions' => 'Enter a valid icon name from <a href="https://feathericons.com">Feather Icons</a> (e.g. `check-circle`).',
```
steffenbew: How about adding target blank? Because if I click it directly while editing, my content changes will be lost. Or is WordPress warning anyways and we'll leave it to the user?

inc/fieldVariables.php
steffenbew: How about adding target blank? Because if I click it directly while editing, my content changes will be lost. Or is WordPress warning anyways and we'll leave it to the user?

emcarru: You'll be warned, so I'd be in favour of leaving it to the user.

PR #174, opened 2019-11-11
feat(FeatureGoogleAnalytics): Removed unused options and modified to …

MyEmpireOfDuctTape:
…fit opt in


Components/BlockCookieNotice/functions.php
```diff
@@ -1,10 +1,20 @@
...
 
 Options::addGlobal('BlockCookieNotice', [
+    [
+        'label' => 'Cookie notice is enabled',
+        'name' => 'cookieNoticeIsEnabled',
+        'type' => 'true_false',
+        'default_value' => 1,
+        'ui' => 1,
+        'ui_on_text' => 'Is Enabled',
+        'ui_off_text' => 'Isn\'t Enabled',
```
steffenbew: I would like to propose different label and values:
Label: "Component Status"
Values: "Activated / Deactivated"

Components/FeatureGoogleAnalytics/README.md
```diff
@@ -1,3 +1,10 @@
...
 
-If a Google Analytics ID is provided in Global Options, the tracking code will be added to all pages, expect for logged in WordPress admins. Entering "debug" instead will log to the browser console only.
+Enable the Google Analytics tracking configuration in the WordPress back end.
+
+If a valid Google Analytics ID is provided, the tracking code will automatically be added to all pages.
+
+The following options are provided under the **"Global Options"** menu:
+
+- Google Analytics ID: This must be a valid Google Analytics ID.
+- Anonymize IP: Check to enable the Google Analytics `_anonymizeIp` function and anonymize the last digits of the user's IP.
```
steffenbew: Please keep the shorter readme (I just updated it 2 days ago). But could you please fix my typo: "expect -> except"?

Components/FeatureGoogleAnalytics/functions.php
```diff
@@ -2,21 +2,63 @@
...
-{
+add_filter('Flynt/addComponentData?name=FeatureGoogleAnalytics', function ($data) {
     $googleAnalyticsOptions = Options::getGlobal('GoogleAnalytics');
+
     if ($googleAnalyticsOptions) {
-        new GoogleAnalytics($googleAnalyticsOptions);
+        $data['jsonData'] = json_encode([
+            'gaId' => $googleAnalyticsOptions['gaId'],
+            'anonymizeIp' => $googleAnalyticsOptions['anonymizeIp'],
+            'serverSideTrackingEnabled' => isTrackingEnabled($googleAnalyticsOptions['gaId']),
```
steffenbew: How about using "isTrackingEnabled' as a variable name, according to the function name? While looking into that function, I'm wondering if you removed the debug mode?

MyEmpireOfDuctTape: I've changed the variable to be called 'isTrackingEnabled' now. I also removed the debug mode from the function. Was this not necessary?

steffenbew: The debug mode should still be there.

Components/BlockCookieNotice/functions.php
MyEmpireOfDuctTape: I changed the label values as specified.

Components/FeatureGoogleAnalytics/README.md
MyEmpireOfDuctTape: I added the readme from the previous version and fixed the typo.

Components/FeatureGoogleAnalytics/functions.php
MyEmpireOfDuctTape: I added the debug mode back now. It's only function would be to see what is pushed to the data layer, which it now logs instead, if the GA ID is set to 'debug'

MyEmpireOfDuctTape:
@domtra @steffenbew I've now merged it with the current master which uses the new fieldvariables.

PR #178, opened 2019-11-13
feat(NavigationFooter): add new layout

emcarru:


Components/NavigationFooter/functions.php
```diff
@@ -16,3 +17,12 @@
 
     return $data;
 });
+
+Options::addTranslatable('NavigationFooter', [
+    [
+        'label' => 'Copyright Text',
+        'name' => 'copyrightText',
+        'type' => 'text',
+        'default_value' => 'Copyright © 2019. All rights reserved'
```
steffenbew: This might be anything, doesn't need to be a copyright notice. I think on our website, I would rather replace it with a notice "Powered by bleech.de" and link our website. So I was thinking about a wysiwyg and a more neutral label and name.

Components/NavigationFooter/_style.scss
```diff
@@ -1,6 +1,24 @@
 [is='flynt-navigation-footer'] {
-  .navigation {
-    padding: $component-spacing-mobile 0;
+  font-size: 0.75rem;
```
steffenbew: `<small>`, `.pill` and `.button--small` are using `0.8rem`. That's why I would propose to use `0.8rem` here as well (or change the BaseStyle).

Components/NavigationFooter/_style.scss
```diff
@@ -1,6 +1,24 @@
...
+
+  .container {
+    padding-bottom: $component-spacing-mobile;
+    padding-top: $component-spacing-mobile;
+    text-align: center;
+
+    @media (min-width: $breakpoint-tablet) {
+      display: flex;
+      flex-direction: row-reverse;
+      justify-content: space-between;
```
steffenbew: If there are no nav items, the copyright notice will move to the right. Minor to me, as there "should" be nav items, but just saying that this might be unintentional.

emcarru: Yeah I didn't want to add more complexity to it when this anyway *shouldn't* happen.

Components/NavigationFooter/functions.php
emcarru: Wysiwyg felt weird to me because of then having the ability to throw headings and lists into here but I didn't think about links. I will update it!

emcarru: I will use the 'basic' toolbar, just remembered we have this

PR #179, opened 2019-11-13
Get fields

domtra:


emcarru:
There's quite some references to using `registerFields` in the docs, but I can strip this all out if we merge this.

Components/BlockCollapse/functions.php
```diff
@@ -2,10 +2,9 @@
 
 namespace Flynt\Components\BlockCollapse;
 
-use Flynt\Api;
-
-Api::registerFields('BlockCollapse', [
-    'layout' => [
+function getLayout()
```
steffenbew: 'layout' as a name makes sense to me when used in a pure ACF context.
As a function name on a component I would rather expect something like `getFields`.

domtra:
@emcarru yes, there is. i intentionally left this open for you to take care of, if we merge this.

Components/BlockCollapse/functions.php
domtra: This is strictly for getting the layout. you might want to create other functions to get other fields. that is why i gave it a specific name, not a general one. it does make more sense to me like this. `getFields` could be any kind of field, or an array of all fields where is not really clear which ones.

steffenbew: I understand that there might be different field groups. But I don't understand why this array of fields would be the layout of the component. Reading the function call `Components\BlockTextImageCrop\getLayout()`, I expect to get "a layout" returned. And then my head goes: "What is the layout of a component in PHP?" and can't make any sense of it. So to me, it's really just about the lacking context that I believe is currently more implicit.

Comparing to the current call `Api::loadFields('BlockTextImageCrop', 'layout')`, I could also imagine something like `Components\BlockTextImageCrop\getFields('layout')` or `Components\BlockTextImageCrop\getACFLayout()` or `Components\BlockTextImageCrop\getLayoutFields()`.

domtra: i would prefer the `getACFLayout` then. i can see what you mean. the thing is, `getLayout` does not return an array of fields, but one layout field where the sub_fields are what we see as fields of the component.

steffenbew: As long as it's specific and a recurring concept (like naming all ACF related functions "getACFSomething"), it's fine with me.

domtra: @steffenbew updated

PR #180, opened 2019-11-13
Esm bundle

domtra:


domtra:
The ideas in this pr are still very relevant. However, those are not top priorities at the moment. Also a lot of code has changed since this was first implemented.
For a final solution we need a better way how to deal with external dependencies and how to seamlessly integrate the multiple bundles into WordPress and our workflow.

PR #181, opened 2019-11-14
docs(README.md): update based on recent changes

emcarru:
- Remove API references
- Remove render flexible component
- Update Options section
- Update table of contents
- Fix some dead links

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@
   * [Advanced Custom Fields](#advanced-custom-fields)
   * [Field Groups](#field-groups)
   * [ACF Option Pages](#acf-option-pages)
-  * [WPML](#wpml)
 * [Maintainers](#maintainers)
 * [Contributing](#contributing)
 * [License](#license)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -152,31 +150,37 @@ add_action('Flynt/afterRegisterComponents', function () {
...
                     'operator' => '==',
-                    'value' => 'page',
+                    'value' => 'page'
                 ],
-            ],
-        ],
+                [
+                    'param' => 'page_type',
+                    'operator' => '!=',
+                    'value' => 'posts_page'
```
steffenbew: What do you think about removing the whole location part, just to keep it short? Or replace it with [...] or something?

steffenbew: @emcarru What do you think about removing the whole location part, just to keep it short? Or replace it with [...] or something?

emcarru: I wanted to have one full example that you could in theory copy and paste and it would work. I could make it shorter by removing the second argument though (so just keep post_type == page). What do you think?

steffenbew: Ok, get it. I thought that "real" examples are in the code already. But fine to keep it the way it is if you prefer it.

PR #193, opened 2019-11-22
ACF Performance

domtra:
feat(AcfPerformance): store group, repeater and flexible content fields as json in db

This will massively reduce the amount of postmeta and options rows needed to store acf group, repeater and flexible content fields in the db.
It does need some more testing, thus draft pr.

### DIFF
--- a/inc/acfPerformance.php
+++ b/inc/acfPerformance.php
@@ -1,0 +1,172 @@
+<?php
+
+namespace Flynt\AcfPerformance;
+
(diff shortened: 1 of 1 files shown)


domtra:
This pr brings drastic improvements for large websites using ACF. I would love to see something like this get integrated into ACF itself. Using this on production sites brings the risks of not being able to update ACF or other plugins not working well. So for now, this will be closed.

PR #195, opened 2019-11-25
feat(ScriptLoader): add async/defer options

domtra:


domtra:
this is this api that will probably eventually make it into core. also i thought about removing the asset register and enqueue methods. we are not really using this. cdn functionality is not gdpr compatible and besides this it does not give much extra. what do you think?

Qakulukiam:
@domtra agreed, we don't really need the register and enqueue (and ofc the internal `Asset::add`) anymore since there are very few files actually enqueued now. If we get rid of those, I am fine with this async / defer update.

steffenbew:
I would love to see that. However, removing `Asset::add` would be a breaking change that we shouldn't need to make.

domtra:
@steffenbew we can always deprecate a function. and only remove it when we decide to release the next major version.

PR #196, opened 2019-11-26
fix(theme): disable big image size threshold

domtra:
this is a quick fix for the new image scaling functionality in wordpress 5.3.

this will make every image work that was uploaded after this fix is deployed. for images that where uploaded before and when on 5.3 we need a bigger fix.

should we go ahead and merge this and think about a more general way afterwards?


steffenbew:
Yes, let's focus on future image uploads and not include a fallback for uploaded images. This could be a separate drop-in script that's added on a project basis.

PR #197, opened 2019-11-26
QA fixes

steffenbew:
Several IE11, Edge fixes (and 1 Firefox) and HTML validation fixes.

### DIFF
--- a/Components/BlockPostFooter/_style.scss
+++ b/Components/BlockPostFooter/_style.scss
@@ -23,6 +23,7 @@
 
     &-info {
       display: flex;
+      flex: 1;
       flex-direction: column;
       justify-content: center;
       padding-left: 1rem;
--- a/Components/GridPostsArchive/index.twig
+++ b/Components/GridPostsArchive/index.twig
@@ -32,7 +32,7 @@
             <h3 class="h3 post-title">{{ post.title }}</h3>
             <p class="post-excerpt">{{ post.preview.read_more(false) }}</p>
             <a class="button button--link" href="{{ post.link }}" title="{{ post.title }}">{{ labels.readMore }}</a>
-          </p>
+          </div>
         </li>
       {% else %}
         <p class="posts-empty">{{ labels.noPostsFound }}</p>
@@ -43,3 +43,4 @@
       {% include 'Partials/_paginationButtons.twig' %}
     {% endif %}
   </div>
+</div>
--- a/Components/ListComponents/_style.scss
+++ b/Components/ListComponents/_style.scss
@@ -27,10 +27,14 @@
       position: relative;
       width: 100%;
 
-      img {
-        min-height: 100%;
-        object-fit: contain;
+      .image {
         position: absolute;
+
+        &--isSmallerThanWrapper {
+          font-family: 'object-fit: contain';
+          height: 100%;
+          object-fit: contain;
+        }
       }
     }
 
--- a/Components/ListComponents/index.twig
+++ b/Components/ListComponents/index.twig
@@ -17,7 +17,7 @@
               {% if componentBlock.componentScreenshot %}
                 <div class="component-image">
                   <div class="component-imageWrapper">
-                    <img class="lazyload"
+                    <img class="lazyload image {{ componentBlock.componentScreenshot.aspect > (1280/800) ? 'image--isSmallerThanWrapper' }} "
                         src="{{ placeholderImage(600, 600 / componentBlock.componentScreenshot.aspect) }}"
                         data-src="{{ componentBlock.componentScreenshot.src }}"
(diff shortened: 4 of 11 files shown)


Components/BlockPostFooter/_style.scss
```diff
@@ -23,6 +23,7 @@
 
     &-info {
       display: flex;
+      flex: 1;
```
Qakulukiam: should maybe be explicitly `flex: 1 1 auto;` to prevent any IE11 issues with the third parameter for flex-basis

steffenbew: Thanks, I'm not familiar with this, but I added `flex: 1` to specifically fix a bug with IE11 where the author text wasn't wrapping, but overflowing the outer container.

steffenbew:
@Qakulukiam I changed `:focus` to `:active` to prevent the state staying active after having opened a link in a new tab. This doesn't effect the accessibility in my opinion, as tabbing through the site still gives focus outlines on elements.

emcarru:
About `:focus` vs `:active`, I think it's important that the browser maintains the focus state after click. It makes sure the user can be aware of the element they interacted with, especially in cases where they unknowingly click on a link that opens in a new tab and then need to switch tabs back to the original website. If you're using a keyboard it's not going to be easy to figure out where you're focused anymore when you switch back if you don't have the `:focus` style.

Components/BlockPostFooter/_style.scss
Qakulukiam: Not sure if it applies here, but here is an example of what I mean: https://makandracards.com/makandra/54644-do-not-use-flex-1-or-flex-basis-0-inside-flex-direction-column-when-you-need-to-support-ie11

Qakulukiam:
In either case, I don't think there is a benefit of using the `:active` selector at all here since it doesn't differ from the hover styles. The hover applies on touch anyways, if that was a concern, and the keyboard `:active` state does not show for triggered links either.

Nice summary here: https://bitsofco.de/when-do-the-hover-focus-and-active-pseudo-classes-apply/

steffenbew:
Thanks @Qakulukiam for your research and input.

@emcarru I feel the `:visited` state would be much more suitable to show which content links you've clicked previously. Especially in the main and footer navigation, I find it confusing to have 2 links visually active when going back to a tab, as it's not clear anymore which site I'm on. Also, browsers don't seem to apply the `:focus` user agent styles themselves, so we're not removing an accessibility feature (but please prove me wrong on this if I am). I would rather have a specific `:focus` style then that gets applied, than having the same as `:hover` and / or active page link.

So should we remove the `:active` and `:focus` state for all links and buttons, as it doesn't provide any benefit, nor does it fix a visual glitch?

emcarru:
Keeping the `:focus` styles is a nice feature because it ensures when you come back to the tab you see exactly which element last had the focus. This also usually happens if you decide to open a link in a new tab yourself (e.g. right click -> open in new tab). You will now have the focus styles until you click/tab away. It only feels like a visual bug in this case because target=blank forces the user into the new tab and then they need to come back. If you were to completely remove `:focus` and only keep `:hover` then you would not have this feature anymore. 

For me it's actually two points wrapped into one:
- "Forcing the user into a new tab makes the `:focus` state looks strange when they come back." I'm personally not a fan of and think this should in 99% cases be left to the user (I love [this article](https://css-tricks.com/use-target_blank/)). Removing target=blank would both let the user decide and also resolve this feeling.
- "Focus states are the same as hover states". I think this is better than nothing, even better: I would love if we actually could add some new, really nice and clear focus styles. When I think of really nice accessible websites like [gov.uk](https://www.gov.uk/) I see they have extra really clear focus styles defined. I'm also a fan of the focus link style on [Smashing Mag](https://www.smashingmagazine.com/). :-)

emcarru:
Just adding an interesting article as now I'm really into this topic 💯 https://zellwk.com/blog/style-hover-focus-active-states/

steffenbew:
Okay, this seems to be a bigger topic than I expected. How about we remove all states besides `:hover` for now (to rely on browser defaults) and create a separate ticket to focus on a solid solution for better `:focus` states?

Components/BlockPostFooter/_style.scss
steffenbew: I tried the suggested fixes. However, they only end up breaking it in IE11 again. Any other suggestions? Otherwise, I would keep it the way it is.

emcarru:
I'd be happier to keep :hover and :focus the same for now and agree on the rest. 😄

steffenbew:
Okay, I'll revert the commit and will make you Lead of the ticket (without ETA for now).

PR #199, opened 2019-11-27
Dynamic resize WPML

domtra:


inc/timberDynamicResize.php
```diff
@@ -44,7 +44,21 @@ function getTableName()
 function getRelativeUploadDir()
 {
     $uploadDir = wp_upload_dir();
-    return $uploadDir['relative'];
+    $homeUrl = getCleanHomeUrl();
+    $relative = str_replace($homeUrl, '', $uploadDir['baseurl']);
+    return $relative;
+}
+
+function getCleanHomeUrl($slug = '')
```
steffenbew: I think a comment here would be nice, explaining that this is a WPML specific fix.

PR #210, opened 2019-12-05
Link and Script type attributes

vanaf1979:
Theme.php
- Add theme support for Html5 style and script tags

CleanHead.php
- Remove id, class, type and media attributes from link tags.
- Remove type attribute from script tags if WordPress is lower than 5.3.
- Added \ to core and native functions to indicate the correct namespace.

Im not sure if these changes have to go to the flynt or flynt-starter-theme  repo!?
If it has to go into the flynt-starter-theme repo, let me know.

### DIFF
--- a/inc/cleanHead.php
+++ b/inc/cleanHead.php
@@ -56,18 +56,30 @@ add_action('init', function () {
 });
 
 /**
- * Remove id, class, type and media attributes from link tags.
+ * cleanStyleTag
+ *
+ * Clean up output of stylesheet <link> tags
  */
-\add_filter('style_loader_tag', function ($input) {
-    return \preg_replace(array( "#\s(id|class|type|media)='[^']+'#", '/\s{2,}/'), array('', ' '), $input);
+add_filter('style_loader_tag', function ($input) {
+    preg_match_all(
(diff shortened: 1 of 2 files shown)


domtra:
Hi @vanaf1979, thanks a lot for you pull request. And for the article explaining the new functionality
I have some remarks. Not sure if you want to update the PR yourself or not.

We want to keep a consistent codebase in terms of coding style. While PSR-2 provides a good base, some things are not covered by it and we have not added custom rules to the linting settings.
Here are styling changes I would like to suggest:
- Do not prefix global WordPress functions with `\`. While this makes a lot of sense, we do not do it anywhere else.
- use `[]` instead of `array()`

On a more general note, really like that you also fixed the script tag clean up for older versions. This is definitely useful. I would not modify the style tag cleanup, however. This introduces some more cleanup that people might rely on. Especially removing `media` could lead to unwanted effect. It makes sense to remove an empty media attribute or `all`, but something like print you would definitely want to stay in there.

vanaf1979:
Hey @domtra,
I adjusted the Pr to your suggestions.
- Fixed the coding style.
- Removed backslash from function calls.
- Reverted back to original style_loader_tag action.

domtra:
hey @vanaf1979, we decided to only add this functionality for newer wordpress versions. this way, we do not change the behaviour for older versions either.
thanks a lot again for contributing.

PR #220, opened 2020-01-19
Validate composer configuration

szepeviktor:
Currently the lock file is out of sync and version should not be in the json file.

kép


szepeviktor:
How about merging this one?

domtra:
hi @szepeviktor, thanks for this. i adjusted the composer.json as well in order to get rid of the warning.

PR #222, opened 2020-01-19
Follow modern coding standard

szepeviktor:
PSR2 is ancient.

### DIFF
--- a/Components/FeatureAdminComponentScreenshots/functions.php
+++ b/Components/FeatureAdminComponentScreenshots/functions.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Flynt\Components\FeatureAdminComponentScreenshots;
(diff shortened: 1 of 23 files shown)


szepeviktor:
If you would like to benefit from WPCs but still follow PSR-12 take a look at this: https://github.com/szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset

domtra:
@szepeviktor thanks for this. i also changed the formatting, so that the linter does not complain.
also thanks for the psr-12 plus wpcs ruleset. we actually plan to move to wpcs completely. this way we do not have to maintain anything extra. also, since flynt is a theme, it makes sense to just follow the wordpress styleguide.

szepeviktor:
You're welcome.

PR #223, opened 2020-01-19
Add syntax and coding style check to CI

szepeviktor:
The syntax check 👹 could be replaced by jakub-onderka/php-parallel-lint


szepeviktor:
For PHPCS config file name see https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file

PR #227, opened 2020-01-27
fix(NavigationFooterStickyFlexBox): made footer sticky to the bottom

AshantiCode:


functions.php
```diff
@@ -18,3 +18,7 @@
     add_action('after_setup_theme', ['Flynt\Init', 'initTheme']);
     add_action('after_setup_theme', ['Flynt\Init', 'loadComponents'], 101);
 }
+
+// Admin Bar.This line disables the admin-bar CSS from pushing the site down. Instead
+// padding was added to the .pageWrapper in base.scss
+add_theme_support('admin-bar', array('callback' => '__return_false'));
```
steffenbew: What have been the challenges with the standard admin-bar CSS?

assets/styles/_base.scss
```diff
@@ -29,6 +31,7 @@ html {
 }
 
 body {
+  height: 100vh;
```
steffenbew: This creates an issue where the body is too large on iOS devices and thus scrollable. Using `height: 100%` instead here and `height` instead of `min-height` on html should fix that.

Just like we've had it before: https://github.com/flyntwp/flynt/commit/b75dc78fb385fca081fefa692e350920d74fb8f0 (In this commit we removed the sticky footer because we thought it was too specific, but it was working well that way.)

assets/styles/_base.scss
```diff
@@ -536,6 +539,25 @@ select {
...
+
+.pageWrapper {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+
+  .admin-bar & {
+    padding-top: 46px;
+
+    @media (min-width: $breakpoint-tablet + 14) {
```
steffenbew: Please change the value to px, as it doesn't relate to our tablet breakpoint.

steffenbew: Only needed if we keep the custom approach instead of the default WordPress margin approach.

PR #230, opened 2020-01-28
feat(BaseStyle):Implement BaseStyle for buttons with right arrow icons

AshantiCode:


assets/styles/_base.scss
```diff
@@ -859,6 +871,16 @@ select {
...
         color: $color-background-dark;
       }
+
+      .feather {
+
+        &:hover,
+        &:focus {
+          background-color: transparent;
+        border-color: $color-background-dark;
+        color: $color-background-dark;
```
steffenbew: Please review indentation.

assets/styles/_base.scss
```diff
@@ -711,6 +718,11 @@ select {
       }
     }
 
+    // &--icon {
+    //    color: white;
+
+    // }
+
```
steffenbew: Please removed unused code.

steffenbew:
Thanks @AshantiCode, an issue I'm seeing is that icons would need to be treated differently depending on their position within the button (left or right). To solve that, I assume we would need extra button classes like `.button--iconLeft` and `.button--iconRight`. However, I only see few specific use cases for that within the components while it would add a good amount of code and complexity for the user as it might end up as an wysiwyg style format as well. I feel the tradeoff would be too big as of now. Might be picked up later though!

PR #231, opened 2020-01-29
feat(GridImageText): Add link option, added responsiveness

AshantiCode:


steffenbew:
@AshantiCode and me decided that we don't like this approach we tried from https://inclusive-components.design/cards/ as it doesn't allow to conveniently open links in new tabs (neither with right click, nor with cmd + click). Might pick up this topic again at some point on a general level to find a unified approach for all clickable cards.

PR #235, opened 2020-02-06
Build and publish with rsync and SSH

pixelsaft:
• Added two variables config > rsyncSrc, rsyncDest
• Added two scripts > publish, publish:rsync

publish:rsync (using rsync to sync minimal required theme files from rsyncSrc to rsyncDest)
publish (build and rsync minimal required theme files to rsyncDest)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,10 @@
   "repository": "git@github.com:flyntwp/flynt.git",
   "author": "bleech <hello@bleech.de>",
   "license": "MIT",
+  "config": {
+    "rsyncSrc": "./",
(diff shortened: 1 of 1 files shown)


domtra:
@pixelsaft thanks for this pr. 
at this point we unfortunately do not want to include anything that is related to deployment in flynt. there a many different ways to deploy a theme. using rsync is one of them. but we do not what to make the decision, what deployment to use for the developer.
i would suggest that you just make these changes, whenever you start a new theme with flynt, or that you add a bash file with the command instead.

PR #236, opened 2020-02-18
Ashanti/optimize print styles

AshantiCode:
I did this without @media only screen, unfortunately i lost my notes about it. But this might be a good global approach.

### DIFF
--- a/assets/admin.scss
+++ b/assets/admin.scss
@@ -5,4 +5,3 @@
 // stylelint-disable-next-line scss/at-import-no-partial-leading-underscore, scss/at-import-partial-extension-blacklist
 @import '../Components/**/_admin.scss';
 
-@import 'styles/print';
--- a/assets/admin.scss
+++ b/assets/admin.scss
@@ -4,4 +4,3 @@
 
 // stylelint-disable-next-line scss/at-import-no-partial-leading-underscore, scss/at-import-partial-extension-blacklist
 @import '../Components/**/_admin.scss';
-
--- a/assets/admin.scss
+++ b/assets/admin.scss
@@ -4,3 +4,5 @@
 
 // stylelint-disable-next-line scss/at-import-no-partial-leading-underscore, scss/at-import-partial-extension-blacklist
 @import '../Components/**/_admin.scss';
+
+@import 'styles/print';
--- a/assets/styles/_print.scss
+++ b/assets/styles/_print.scss
@@ -1,0 +1,61 @@
+@media print {
+  [is='flynt-navigation-main'],
+  [is='flynt-navigation-burger'],
+  [is='flynt-navigation-footer'],
+  [is="flynt-block-cookie-notice"],
+  .button,
+   {
+    display: none !important;
+  }
+
+  a:after {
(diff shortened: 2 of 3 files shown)


assets/admin.scss
```diff
@@ -4,3 +4,5 @@
 
 // stylelint-disable-next-line scss/at-import-no-partial-leading-underscore, scss/at-import-partial-extension-blacklist
 @import '../Components/**/_admin.scss';
+
+@import 'styles/print';
```
steffenbew: I don't think we need to have the print styles in the admin sheet.

assets/styles/_print.scss
```diff
@@ -0,0 +1,84 @@
...
+    box-shadow: none !important;
+    text-shadow: none !important;
+  }
+
+  [is='flynt-navigation-main'],
+  [is='flynt-navigation-burger'],
+  [is='flynt-navigation-footer'],
+  [is="flynt-block-cookie-notice"],
+  .button,
+  footer {
```
steffenbew: Instead of hiding individual components, I would suggest to hide the wrapping sections of the layout: `.mainHeader`, `.mainFooter`, `.wpFooter`.
Also, I think buttons should still be visible and not every `footer` element should be hidden, as it might be used differently in a context where you want it to be visible.

assets/styles/_print.scss
```diff
@@ -0,0 +1,84 @@
...
+    display: none !important;
+  }
+
+  a {
```
steffenbew: I noticed this one is not having the `:visited` state that the HTML5 Boilerplate suggests. Why?

assets/styles/_print.scss
```diff
@@ -0,0 +1,84 @@
...
+    text-decoration: underline;
+    font-weight: bold;
```
steffenbew: Why make links bold? I believe they should look like in the design, but black and underlined.

assets/styles/_print.scss
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  a[href]:after {
+    content: " (" attr(href) ")";
+  }
+
+  a[href^="/"]:after {
```
steffenbew: I assume this would not show the urls of internal links, right? Why should we hide those?

assets/styles/_print.scss
AshantiCode: I thought that this rule for the <a> tag will apply to every <a> tag on the page, if `a:visited ` will get underlined, it will anyways get underlined, cause it is an <a> tag. But obviously my thoughts were wrong due to your question.

assets/styles/_print.scss
AshantiCode: okay,removed it.

assets/styles/_print.scss
AshantiCode: Was something I found when searching for print styles on the web. I removed it.

AshantiCode:
I added some rules to the `.component-footer` of ListComponents otherwise the link addresses weren't readable and would overlap.

PR #255, opened 2020-06-02
fix(baseStyle): move duplicate variables to variables file

koukouchris:
and fix data:image missing parenthesis causing IE bug on hover:
Screen Shot 2020-06-02 at 18 06 28

### DIFF
--- a/assets/styles/_base.scss
+++ b/assets/styles/_base.scss
@@ -352,6 +352,9 @@ button {
     }
 
     &::after {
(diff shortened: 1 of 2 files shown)


steffenbew:
I understand that you extracted the rgb variables into the global `_variables.scss` to not repeat the variable definitions. In my opinion, these variables are not defining new values, but are just helpers to extract the rgb values from global variables. This extraction is very component specific, that's why keeping the variables local makes sense to me.

It would be great to have the missing parenthesis fix merged individually.

PR #258, opened 2020-06-08
Mykola/css variables

MadFlip:


assets/styles/_variables.scss
```diff
@@ -26,5 +26,33 @@ $color-background-light: #f2f6fe;
 $color-background-dark: #091a41;
 $color-background-hero: $color-brand;
 
+:root {
+  // editable colors for WP Customizer
+  --color-brand: #{$color-brand};
+  --color-accent: #{$color-accent};
+  --color-accent-dark: #{darken($color-accent, 10%)};
```
koukouchris, line 33: This variable is only used once from what i see in `.themeReset`, I don't think it should be part of editable colours, it seems connected to `color-accent` so it should change accordingly with customizer changing `color-accent`. Is this possible?

assets/styles/_variables.scss
```diff
@@ -26,5 +26,33 @@ $color-background-light: #f2f6fe;
...
+  --color-contrast: #{$color-contrast};
```
koukouchris, line 34: This colour is used in different elements, but also in `color-text`. Seems a better approach to be able to change just the color of the text in customizer. I can imagine wanting to change all texts colour to black for example, what do you think? Should we create a new variable for color-text?

assets/styles/_variables.scss
```diff
@@ -26,5 +26,33 @@ $color-background-light: #f2f6fe;
...
+  --color-headline: #{$color-headline};
+  --color-border: #{$color-border};
+  --color-background-light: #{$color-background-light};
+  --color-background-dark: #{$color-background-dark};
+  // end of editable colors
+
+  --color-background: #{$color-background}; // TODO: should it be editable color?
```
koukouchris, line 41: Since its used in a lot of elements right now would be confusing on what changes exactly if we add it to customizer

assets/styles/_variables.scss
MadFlip: **--color-accent-dark** is a link `hover` color. I tried to generate darken color based on **--color-accent** but scss function doesn't work with css variables. The only way I found is using color in hsl, using hue values as variable and change lightness to make it darker:
```
--color-accent-hue: 20;
--color-accent: hsl(var(--color-accent-hue),  95%,  **53%**);
--color-accent-dark: hsl(var(--color-accent-hue),  95%,  **43%**);
```
but ie11 polyfill doesn't work with such solution :(
or maybe we can change link hover color to `--color-background-dark` to make it consistent with button link and remove --color-accent-dark at all
image

assets/styles/_variables.scss
MadFlip: image
**--color-text** already exists and by default it takes **--color-contrast** color. You can put **--color-text** into customizer instead of **--color-contrast**, or use **--color-contrast** by name label as Text color

assets/styles/_variables.scss
MadFlip: I agree, don't add it to the customizer for now. We can test it later, if some element will need to change white color we can add new variable for that, something like --color-white

inc/customizer-colors.php
```diff
@@ -0,0 +1,171 @@
...
+        $wp_customize,
+        'theme_colors_light_theme',
+        [
+            'section' => 'theme_colors',
+            'label'   => esc_html__('Light Theme', 'flynt'),
+            'description' => 'Changes light theme background color, table row and caption background.'
+        ]
+    ));
+
+    // Theme Dark color
```
steffenbew, line 115: "Dark" should be lowercase to be consistent

inc/customizer-colors.php
```diff
@@ -0,0 +1,171 @@
...
+        $wp_customize,
+        'theme_colors_headline',
+        [
+            'section' => 'theme_colors',
+            'label'   => esc_html__('Headline', 'flynt'),
+            'description' => 'Changes color of all headlines and form elements.'
+        ]
+    ));
+
+    // Theme Hero color
```
steffenbew, line 77: "Hero" should be lowercase to be consistent

assets/styles/_base.scss
```diff
@@ -481,6 +464,32 @@ select {
...
+    z-index: 1;
+  }
+
+  select {
+    // don't change backgound, if you need to change it use wrapper .select
+    background: none;
+    position: relative;
+    z-index: 2;
+  }
+}
```
steffenbew, line 491: I assume that this is enabling the small icon to take the color from the customizer. Looks like a great solution! Did you check if it works with the Contact Form 7 form? Does the form component need changes?

assets/styles/_variables.scss
```diff
@@ -14,17 +14,32 @@ $component-spacing-tablet: 80px;
...
-$color-background: #fff;
-$color-contrast: #414751;
-$color-text: $color-contrast;
-$color-headline: #0b1016;
-$color-border: #d8d8d8;
+:root {
+  --color-border: #d8d8d8; // TODO: Add this on customizer?
+  --color-background: #fff;
+  --color-contrast: #414751;
+  --color-background-hero: var(--color-brand);
```
steffenbew, line 21: I believe this is also an editable color and should move down.

assets/styles/_variables.scss
```diff
@@ -14,17 +14,32 @@ $component-spacing-tablet: 80px;
...
-$color-accent: #f96417;
-$color-background: #fff;
-$color-contrast: #414751;
-$color-text: $color-contrast;
-$color-headline: #0b1016;
-$color-border: #d8d8d8;
+:root {
+  --color-border: #d8d8d8; // TODO: Add this on customizer?
+  --color-background: #fff;
+  --color-contrast: #414751;
```
steffenbew, line 20: I believe the background and contrast colors should also be editable in the customizer.
We wouldn't need an option to change the text color, as it's the same as the contrast color. Or we could get rid of the contrast color and only use the text color.

assets/styles/_variables.scss
```diff
@@ -14,17 +14,32 @@ $component-spacing-tablet: 80px;
...
 
-$color-brand: #0d8eff;
-$color-accent: #f96417;
-$color-background: #fff;
-$color-contrast: #414751;
-$color-text: $color-contrast;
-$color-headline: #0b1016;
-$color-border: #d8d8d8;
+:root {
+  --color-border: #d8d8d8; // TODO: Add this on customizer?
```
steffenbew, line 18: I think so! :)

assets/styles/_base.scss
MadFlip: yes, it work's with Contact Form 7. The component itself doesn't need any specific changes (just message with forms examples need small adjustment, that I already did and pushed to premium components repo). And In the backend in Contact Forms template we need to wrap select into span with .select class. That's it
image

assets/styles/_variables.scss
```diff
@@ -14,17 +14,31 @@ $component-spacing-tablet: 80px;
...
+  --theme-button-background: var(--color-accent);
+  --theme-button-background-hover: var(--color-background-dark);
+  --theme-button-color: var(--color-background);
+  --theme-button-color-hover: var(--color-background);
+  --theme-link-color: var(--color-accent);
+  --theme-link-color-hover: var(--color-background-dark);
+  --theme-pill-background: var(--color-background-dark);
+  --theme-title-color: var(--color-headline);
+  --theme-border-color: var(--color-border);
+  --theme-transparent: transparent; // needs for ie11-custom-propertiess polyfill to work properly with .themeReset
```
steffenbew, line 40:
```suggestion
  --theme-transparent: transparent; // needed for ie11-custom-properties polyfill to work properly with .themeReset
```

assets/styles/_base.scss
steffenbew: I just had a fresh look on this and have 2 considerations about the change:
1. This changes the default markup for the select element, which needs to be learned by people using Flynt. We should prevent people having to learn new things whenever there is no big advantage to it.
2. Default browser UI elements (like date picker icon) have a dark color and don't change based on the background color. Changing the whole form UI to a dark background, would need other elements to be changed as well to look perfect. However, I expect them to still be usable even with the dark elements.

That's why I propose to stay with the default browser markup and a dark icon for all themes.

MadFlip: select has fallback with dark icon if person forget to wrap select in .select

templates/basestyle.twig
```diff
@@ -200,7 +200,7 @@
           </fieldset>
           <p>The user agent stylesheet typically has a very weird <b>&lt;fieldset&gt;</b> and <b>&lt;legend&gt;</b> styling. But hey: why not customize it to our needs?!</p>
           <fieldset>
-            <legend>Checkbox Inputs</legend>
+            <legend style="display: block;">Checkbox Inputs</legend>
```
steffenbew, line 203: @MadFlip Can we avoid this? Maybe with a `br` tag?

MadFlip: `br` add too big gap, so I wrapped content below legend with `div`. I did this to prevent inputs jump on the same line as `legend` title on IE

MadFlip:
> @MadFlip I assume this can be prevented by putting `overflow: hidden` on the `.boxRadius` container element.

Yes, good advise, thanks!

assets/styles/_base.scss
```diff
@@ -33,6 +33,10 @@ html {
 body {
   height: 100%;
   overflow-x: hidden;
+
+  @media (min-width: $breakpoint-mobile-horizontal) {
+    font-size: 1.13rem;
```
steffenbew, line 38: This translates to 18.08px, it would be great ending up with an even number of 18px instead.
```suggestion
    font-size: 1.125rem;
```

Additionally, I would be curios if we can use that size on mobile as well!

assets/styles/_base.scss
```diff
@@ -263,51 +309,88 @@ button {
...
   font-weight: 700;
   line-height: inherit;
-  padding: 1rem 1.5rem;
+  outline: none;
+  padding: 0.875rem 1.5rem;
   text-align: center;
   text-decoration: none;
-  text-transform: uppercase;
-  transition: background-color 0.2s ease, border-color 0.2s ease, color 0.2s ease;
+  transition: background-color 0.2s ease, border-color 0.2s ease, color 0.2s ease, box-shadow 0.2s ease;
```
steffenbew, line 335: Could we use a shortcut?

```suggestion
  transition: all 0.2s ease;
```

assets/styles/_variables.scss
```diff
@@ -8,23 +8,64 @@ $container-max-width: 1440px;
...
 
-$color-brand: #0d8eff;
-$color-accent: #f96417;
-$color-background: #fff;
-$color-contrast: #414751;
-$color-text: $color-contrast;
-$color-headline: #0b1016;
-$color-border: #d8d8d8;
+// Theming variables
+:root {
```
steffenbew, line 18: As `.themeReset` is supposed to be the same as `:root`, could we combine the selectors?

```suggestion
:root,
.themeReset {
```

inc/customizer-colors.php
```diff
@@ -0,0 +1,413 @@
...
+    //Check if color has 6 or 3 characters and get values
+    if (strlen($color) == 6) {
+            $hex = array( $color[0] . $color[1], $color[2] . $color[3], $color[4] . $color[5] );
+    } elseif (strlen($color) == 3) {
+            $hex = array( $color[0] . $color[0], $color[1] . $color[1], $color[2] . $color[2] );
+    } else {
+            return $default;
+    }
+
+    //Convert hexadec to rgb
```
steffenbew, line 297: Please add a space after `//` here and on other comments.

assets/styles/_base.scss
```diff
@@ -196,55 +222,75 @@ ol {
...
+
+    &:first-child {
+      border-top-left-radius: 6px;
+    }
+
+    &:last-child {
+      border-top-right-radius: 6px;
+    }
+  }
 }
```
steffenbew, line 264: Please test again if this is needed, since adding `overflow: hidden` to `.boxRadius`.

steffenbew:
@domtra Could you please review the PHP parts?

lib/Customizer/InfoControl.php
```diff
@@ -0,0 +1,21 @@
...
+
+class InfoControl extends WP_Customize_Control
+{
+    public $type = 'info';
+    public function render_content() // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    {
+        ?>
+         <?php if ($this->label) : ?>
+            <h2 style="margin: 0 0 10px;" class="panel-title"><?php echo $this->label; ?></h2>
+         <?php endif ?>
```
steffenbew, line 15: Linter complains about indentation being wrong.

assets/styles/_base.scss
MadFlip: `In general, the rule of thumb is that font size needs to be 16 pixels for mobile websites. Anything smaller than that could compromise readability for visually impaired readers. Anything too much larger could also make reading more difficult. You want to find that perfect Goldilocks formula and, time and time again, it comes back to 16 pixels.`

I agree with this statement, I like how 16px looks on mobile. 18px looks too big for me

assets/styles/_base.scss
MadFlip: yes, this is still needed for IE 😕

assets/styles/_base.scss
steffenbew: Alright, thanks for researching this topic!

Components/GridPostsArchive/_style.scss
```diff
@@ -14,33 +13,76 @@
...
-      text-decoration: none;
+      @include reset-link;
+      display: flex;
+      flex-direction: column;
+      height: 100%;
     }
 
     .figure {
-      background-color: $color-text;
+      flex-shrink: 0; // IE11 bug fix (prevent image from stretchin')
```
steffenbew, line 31: I think this can be removed as we're dropping IE11 support. Please also see for other IE11 support that can be removed.

Components/BlockImage/index.twig
```diff
@@ -1,22 +1,23 @@
...
-            {{ image.src|resizeDynamic(1320) }} 1320w,
-            {{ image.src|resizeDynamic(1024) }} 1024w,
-            {{ image.src|resizeDynamic(800) }} 800w,
-            {{ image.src|resizeDynamic(600) }} 600w,
-            {{ image.src|resizeDynamic(375) }} 375w"
-          data-sizes="auto"
-        {% endif %}
-        alt="{{ image.alt|e }}">
+      {% if image.post_mime_type == 'image/svg+xml' or image.post_mime_type == 'image/gif' %}
+        <img src="{{ image.src }}" alt="{{ image.alt|e }}">
```
steffenbew, line 5: Should this also get a lazyload class? I thought it might be just missing in the original code snippet as well.

Components/NavigationBurger/functions.php
```diff
@@ -14,7 +14,7 @@
 add_filter('Flynt/addComponentData?name=NavigationBurger', function ($data) {
     $data['menu'] = new Timber\Menu('navigation_burger');
     $data['logo'] = [
-        'src' => get_theme_mod('custom_header_logo') ? get_theme_mod('custom_header_logo') : Asset::requireUrl('Components/NavigationBurger/Assets/logo.svg'),
+        'src' => get_theme_mod('custom_header_logo') ? get_theme_mod('custom_header_logo') : Asset::requireUrl('../assets/images/logo.svg'),
```
steffenbew, line 17: Do we need the `../` relative path? I expected the folder being on the same level as `Components`.

PR #265, opened 2020-06-25
Block video oembed refactor

koukouchris:


Components/BlockVideoOembed/functions.php
```diff
@@ -57,7 +57,26 @@ function getACFLayout()
...
+                        'type' => 'radio',
+                        'other_choice' => 0,
+                        'save_other_choice' => 0,
+                        'layout' => 'horizontal',
+                        'choices' => [
+                            'sizeSmall' => 'Small',
+                            'sizeMedium' => 'Medium',
+                            'sizeLarge' => 'Large (Default)',
+                            'sizeHuge' => 'Huge',
+                            'sizeFull' => 'Full',
```
steffenbew, line 73: I think these values need to be translated, like over here: https://github.com/flyntwp/flynt/blob/master/Components/BlockImage/functions.php#L44-L55

PR #268, opened 2020-06-26
Image optimization

koukouchris:


Components/GridPostsArchive/index.twig
```diff
@@ -17,12 +17,16 @@
               <figure class="figure">
                 <img class="figure-image lazyload"
                   src="{{ post.thumbnail.src|resizeDynamic(424, 283) }}"
-                  srcset="{{ placeholderImage(424, 283) }}"
+                  srcset="{{ placeholderImage(3, 2) }}"
```
steffenbew, line 20: Could you please elaborate why you changed the placeholder dimension value?
I actually like it with the "real" numbers for 2 reasons:
1. It tells me that the placeholder value is the same size, I don't need to re-calculate.
2. The ratio is a tiny bit off, which might lead to rendering issues due to rounding.

Edit: I just understood that 424:283 is the only value which is not exactly a 3:2 ratio, but it's the actual desktop size. I thus understand why you chose the new 3:2 value. However, I still feel the approach is inconsistent to how we're using the placeholder in other components. But even if we would go with 3:2, the value should be updated in GridPostsLatest as well.

koukouchris: Yes 424:283 is a bit off from 3:2 aspect ratio. Actually, I took this approach from GridPostsLatest and adjusted all grids with the same logic. So now all grid components have the same values.

steffenbew: Ah great. I thought it wasn't updated in GridPostsLatest yet. My fault. All fine then :)

Issue #261, opened 2020-06-23
Remove Gutenberg styles from the head tag


HanifBirgani:
Considering that Flynt disables the Gutenberg by default, we can safely remove Gutenberg styles from the head tag:

```php
/**
 * Remove Gutenberg default styles
 */
add_action('wp_print_styles', function (): void {
    wp_dequeue_style('wp-block-library');
    wp_dequeue_style('wp-block-library-theme');
});
```

I can make a PR to add this to the core.

steffenbew:
True! However, Gutenberg adds even more stuff that we like to disable with a plugin like [Disable Gutenberg](https://wordpress.org/plugins/disable-gutenberg/).

HanifBirgani:
You are right about completely disabling Gutenberg by a plugin, but my suggestion only focuses on front-end codes. I think Flynt could deliver a cleaner head by removing unused Gutenberg front-end styles.

steffenbew:
Right, and I agree it makes sense. I just don't want to integrate something seemingly simple now, that could become a complex feature over time, with even more functionality to disable Gutenberg. Because that's something existing plugins provide already. 
At the same time, we're also removing the editor in the theme already. That's why removing header styles in the frontend seems to be a sensible thing to do.

I would propose to add it to the [removeEditor.php](https://github.com/flyntwp/flynt/blob/master/inc/removeEditor.php) file directly, as it's tied to each other.

What do you think? Would you like to submit a pull request?

PR #270, opened 2020-06-27
Removes Gutenberg default styles on front-end #261

HanifBirgani:
This PR removes Gutenberg default styles on front-end which mentioned in #261.


steffenbew:
@domtra Looking good to me! Any concerns?

domtra:
closes #261

Issue #250, opened 2020-05-27
Restrict access to .twig files


HanifBirgani:
I think it's better to restrict access to .twig files in the webserver config level.

Apache:
```
<FilesMatch ".+\.twig$">
    <IfModule mod_authz_core.c>
        # Apache 2.4
        Require all denied
    </IfModule>
    <IfModule !mod_authz_core.c>
        # Apache 2.2
        Order deny,allow
        Deny from all
    </IfModule>
</FilesMatch>
```

Nginx:
```
location ~- \.twig$ {
  deny all;
}
```

domtra:
Hi @HanifBirgani , thanks for your comment. I agree that it is a good idea to restrict access to the twig files. Usually I would think to leave the configuration of the webserver to the theme user. With php files in the theme folder it is basically the same.
What would you suggest for Flynt, or any wordpress theme? Do you have an example what other themes do? Write to the htaccess or just give advice in wp-admin?

HanifBirgani:
I suggest a .htaccess rewrite rule on theme init.

The sage theme suggested the rules on the [installation page](https://roots.io/docs/sage/9.x/installation/#server-configuration) (for .blade.php files) but did not force users to do it.

HanifBirgani:
Is there any update on this issue?

steffenbew:
@domtra is currently out of office until the end of the week. I'm sure he will look into this again when he's back.

PR #271, opened 2020-06-28
Add rules to .htaccess to restrict access to .twig files

HanifBirgani:
Add access restriction rules to .htaccess for twig files on theme setup. Previously mentioned in #250.


domtra:
Hi @HanifBirgani, thanks for the pull request.

Unfortunately we will not merge this one. We want to include as little code as possible that depends on infrastructure. 
We might include a note in the docs, however, to make sure that people are aware of this.

Thanks for bringing this up again!

domtra:
closes #250

PR #277, opened 2020-07-15
feat(theme): add heading format classes

alexheinrich:
Hi Steffen, I have added the heading format classes. Does this look good to you?

### DIFF
--- a/inc/tinyMce.php
+++ b/inc/tinyMce.php
@@ -88,32 +88,32 @@ function getConfig()
                     [
                         'title' => 'Heading 1',
                         'classes' => 'h1',
-                        'selector' => 'h1,h2,h3,h4,h5,h6'
+                        'selector' => 'h1,h2,h3,h4,h5,h6,p'
                     ],
                     [
                         'title' => 'Heading 2',
                         'classes' => 'h2',
-                        'selector' => 'h1,h2,h3,h4,h5,h6'
+                        'selector' => 'h1,h2,h3,h4,h5,h6,p'
                     ],
                     [
                         'title' => 'Heading 3',
                         'classes' => 'h3',
-                        'selector' => 'h1,h2,h3,h4,h5,h6'
+                        'selector' => 'h1,h2,h3,h4,h5,h6,p'
                     ],
                     [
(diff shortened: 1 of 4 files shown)


inc/tinyMce.php
```diff
@@ -81,6 +81,42 @@ function getConfig()
...
         ],
         'styleformats' => [
+            [
+                'title' => 'Headings',
+                'icon' => '',
+                'items' => [
+                    [
+                        'title' => 'Heading 1',
+                        'classes' => 'h1',
+                        'selector' => 'h1,h2,h3,h4,h5,h6,p'
```
steffenbew, line 91: Looking great! I've got 2 thoughts I would like your opinion about:

1. Why do you prefer a specific selector, instead of allowing the classes for all elements?
2. I we keep the selector, how about removing the selector`h1` for a Heading 1 `h2` for Heading 2, and so on?

alexheinrich: I've allowed the heading classes to be added to anything, since I think that does make sense.
The reason I added all of the selectors was that TinyMCE uses the first selector to style the item in the dropdown menu. The result was that "Heading 1" was smaller than the others in the list... We, of course, add the backend styles so this wouldn't be a big issue, but I thought someone using flynt out of the box might find it strange that the largest heading was smaller than the others. Allowing the classes to be given to all elements bypasses this though.

alexheinrich:
I hope it is ok that I pushed it to this branch but I added some fixes for some minor A11y issues that seem to have slipped through the cracks. Would you mind taking a quick look and see what you think?

steffenbew:
> I hope it is ok that I pushed it to this branch but I added some fixes for some minor A11y issues that seem to have slipped through the cracks. Would you mind taking a quick look and see what you think?

Please add the a11y fixes in a separate branch to make the version notes specific to the changes.

inc/tinyMce.php
steffenbew: Alright, I didn't know that! I agree with your point.

alexheinrich:
I have reverted the commits, so this is ready to merge.

steffenbew:
Great, thanks!

PR #284, opened 2020-08-06
Unify css

koukouchris:


Components/GridPostsArchive/Partials/_paginationButtons.scss
```diff
@@ -23,13 +23,13 @@
     width: 16px;
   }
 
-  &-nextLink {
+  &-nex {
```
steffenbew, line 26: This looks like a typo.

koukouchris: Oh no! Fixed

PR #286, opened 2020-08-11
(fix)NavigationFooter: fallback text added

pixelsaft:
Added a simple fallback text consisting of the current year and site name.


steffenbew:
Thanks for your contribution. I discussed it with the team and we've been split about adding such a fallback or not. We've also discussed alternative solutions to yours, which amplified that content should live within the backend, not within the template. It wouldn't be able for a content manager to remove the copyright notice. That's why I won't accept this change.

PR #297, opened 2020-08-13
feat(NavigationFooter): copyright notice, site title

pixelsaft:
@steffenbew: It would be able for a content manager to remove the copyright notice, now. 

We have added a new Tab "Options" with two switches at the "Translatable Options" from the NavigationFooter (this is not 100% consistent but the Component "Grid Posts Archive" has also an "[Options](https://github.com/flyntwp/flynt/blob/478c3fa2472a781be03d5a17e8479cb2e5335386/Components/GridPostsArchive/functions.php#L149)" tab to setup at this area.


steffenbew:
I see. This would allow to customize it. But it's adding a lot of clutter to the backend UI. And all of that for information that in my opinion doesn't add much value to a website.
Harun from our team suggested that we could add the functionality of placeholder variables instead:
Screen Shot 2020-08-12 at 09 49 34
I like that idea, but I wouldn't be sure about the implications that would come with it. This could also be achieved with [a shortcode](https://css-tricks.com/snippets/wordpress/year-shortcode/) or [a plugin](https://wordpress.org/plugins/current-year-shortcode/). I would still need to think if and how it would make sense to make it a part of Flynt. We should be careful about introducing new features.

pixelsaft:
I think placeholder variables nor a shortcode and a plugin are not very comfortable for content managers, because they must remember that functionality. 

My first idea was to display the copyright notice and the title of the site as simple fallback only if no content is inserted — the main advantage of this functionality is, that the copyright notice will be shown from scratch, nobody must adjust the year manually. And in company sites, the site title often represents the company name.

I‘m still thinking about other ways for implementation, but I’m not sure what is the main target behind to get this feature implemented and which solutions do you prefer.

steffenbew:
I appreciate your effort and thinking. I agree with your point that having to remember shortcodes is an issue. That's why I wouldn't implement it without any other good reasons and solutions that aim to make the experience seamless.
At the same time, I'm not a big fan of these kind of copyright notices, as I think they're legally not necessary. Thus, forcing them on the user, or adding multiple options to handle them sounds like a lot to me.

pixelsaft:
@steffenbew: I understand your position! We have many customers who want display a copyright notice, so that i think it could be a good starting point to implement it as a feature.

I have checked my mind and updated the component again. Hope this is a better way ;)

steffenbew:
@pixelsaft I have had another look at your revisited solution. I like that you thought about different options. On the other side, I could think of even more options how to format such notice. I don't want to create a seemingly complicated logic and many options for something that has a low impact on the overall site.

What I could still imagine is adding a [default_value](https://github.com/flyntwp/flynt/blob/master/Components/BlockNotFound/functions.php#L21) to the wysiwyg, with custom shortcodes:
`© [year] [blogname]` which would resolve into `© 2020 Website Name`.

As the available shortcodes are prepopulated, a content manager knows their options. Additionally, the available shortcodes could be listed in the field `instructions`. The advantage I'm seeing with this approach, is the flexibility in changing the notice into anything else, i.e. by not using `[blogname]` but a custom company name instead, or amending a legal entity, or a copyright claim.

The shortcodes could be defined in new file: `/inc/shortcodes.php`. This new shortcode approach would set a direction for similar use cases and keeps the code base rather clean and simple.

I took the following article as a reference: https://divimundo.com/en/blog/copyright-year-wordpress-footer/

What do you think?

pixelsaft:
@steffenbew all good ones are three – just updated according to your feedback.

steffenbew:
Thanks for the update, @pixelsaft. Looking great! The shortcode examples are also very helpful! I would like to suggest the following changes:

- Please change the shortcode options and examples into something like a table, to give an easier overview. Or just less text and divide it into "Shortcodes" and "Examples"?
- The colors and margins look nice, but they introduce new custom styles. I would prefer to remove custom styles.

On another note, the tab content appears twice in the WordPress backend for me. Is it just me? Can't spot anything wrong with the code.

pixelsaft:
@steffenbew according to your feedback we made some changes. 

- _Please change the shortcode options and examples into something like a table, to give an easier overview. Or just less text and divide it into "Shortcodes" and "Examples"?_ > Done.
- _The colors and margins look nice, but they introduce new custom styles. I would prefer to remove custom styles._ > Done. Please note: The custom styles we have added before are also present at the premium component “FormContactForm7”.

steffenbew:
@pixelsaft Thanks for the changes!
* The table looks nice to me. I actually thought it would be labels on the left, values on the right to give a top to bottom overview. But with 50% wide columns, it would be too large as well to see the relationship of labels and values. I don't want to obsess about it, you probably checked a couple of options and decided for the one that works best.
* You're absolutely right about the styles! I didn't remember we added custom styles in the Contact Form 7 component. But I like it the way you changed it now.
* I noticed the `[tagline]` shortcode also returned the site title for me. I think the syntax of the function call `get_bloginfo(false, 'description');` needs to be the following instead `get_bloginfo('description');`. Same should be changed for the `[sitetitle]` shortcode function call.

steffenbew (approved):

PR #299, opened 2020-08-18
Fix Windows Dynamic Resize

harunbleech:


domtra:
@harunbleech, great. thank you.  
generally i try to avoid using these kind of `str_replace`s. 
do you think you can achieve the same thing with using `realpath()` on `$homePath` and `$uploadDir['basedir']`?

harunbleech:
@domtra using `realpath()` give us `\wp-content\uploads`. Then we would need to remove this first backslash. The rewrite rule will work but it will look like this `^wp-content\uploads/resized/?(.*?)/?$`

harunbleech:
@domtra `get_home_path()` does the same at the end `return str_replace( '\\', '/', $home_path );`

PR #300, opened 2020-08-18
Fix gulp revUpdateReferences on Windows

harunbleech:
Using the `path.join(config.dest, '/**/**.{css,js}')` on Windows a dist folder is created inside dist folder with all the updated references and the `dist/assets/*.{js,css}` files weren't updated with the references.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -91,7 +91,7 @@ function revUpdateReferences (config) {
     const rewrite = require('gulp-rev-rewrite')
     var manifest = gulp.src(path.join(config.dest, 'rev-manifest.json'))
 
-    return gulp.src(path.join(config.dest, '/**/**.{css,js}'))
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -91,7 +91,7 @@ function revUpdateReferences (config) {
     const rewrite = require('gulp-rev-rewrite')
     var manifest = gulp.src(path.join(config.dest, 'rev-manifest.json'))
 
-    return gulp.src(path.join(config.dest, '/**/**.{css,js}'))
```
domtra, line 94: would `return gulp.src(path.join(config.dest, '**/**.{css,js}'))` work as well?

harunbleech: No, it still does the same.

domtra (approved):

PR #304, opened 2020-09-03
fix(SliderImages): add polyfill to fix Swiper v6 in IE

MadFlip:


steffenbew:
Thanks @MadFlip! I don't have much understanding of the package and why it's needed. While I trust you with doing the right thing, I'm not qualified to give an approval. @domtra could you please review this PR?

domtra (approved):

PR #308, opened 2020-09-22
feat(reusableComponents): set up reusable cpt, options, move componen…

koukouchris:
…ts to different file, blockReusable


Components/ReusableComponent/admin.js
```diff
@@ -0,0 +1,20 @@
+/* globals acf */
+import $ from 'jquery'
+
+acf.addAction('select2_init', function ($select, args, settings, field) {
+  const $fieldEl = field.$el
+
+  $select.on('change', { $fieldEl: $fieldEl }, function () {
+    const postId = $(this).val()
+    const postTitle = $(this).find('option:selected').text()
+    const postLink = `/wp/wp-admin/post.php?post=${postId}&action=edit&classic-editor`
```
steffenbew, line 10: I think the admin link could be different based on the installation. The path `wp` could be different. I assume this can be fixed by using WordPress path variables or functions.

Components/ReusableComponent/admin.js
```diff
@@ -0,0 +1,20 @@
...
+    const $postLink = $fieldEl.find('.reusable-postLink')
+
+    if ($postLink.length > 0) {
+      $postLink.text(postTitle)
+      $postLink.attr('href', postLink)
+    } else {
+      $fieldEl.find('.description').html(`Edit <a class="reusable-postLink" href="${postLink}" target="_blank" rel="noopener noreferrer">${postTitle}</a>.`)
+    }
```
steffenbew, line 18: I think it would be nice to keep the information about adding or editing components, and appending the edit link to the selected component to it (or to place it after the select input).

It would als be nice if this string was translatable with WordPress default functions.

Components/ReusableComponent/functions.php
```diff
@@ -31,6 +32,17 @@ function getACFLayout()
...
+        ],
     ];
 }
+
+add_filter('acf/prepare_field/name=reusableId', function ($field) {
+    // Update instructions with selected post link
+    if ($field['value']) {
+        $postTitle = get_the_title($field['value']);
+        $instructions = $field['instructions'];
+        $instructions = 'Edit <a class="reusable-postLink" href="/wp/wp-admin/post.php?post=' . $field['value'] . '&action=edit&classic-editor" target="_blank" rel="noopener noreferrer">' . $postTitle . '</a>.';
```
steffenbew, line 44: Looks like the first `$instructions` definition isn't needed, as it's overwritten right after.

Do we also need the parameter `&classic-editor` in the link?

Components/ReusableComponent/functions.php
```diff
@@ -31,6 +32,17 @@ function getACFLayout()
                 'return_format' => 'id',
                 'instructions' => 'Add or edit <a href="/wp/wp-admin/edit.php?post_type=reusable-component">reusable components</a>.'
```
steffenbew, line 33: I don't seem to be able to deselect a chosen component. I could imagine it being helpful.

steffenbew (changes requested):

Components/ReusableComponent/functions.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    return $data;
+});
+
+function getACFLayout()
+{
+    return [
+        'name' => 'ReusableComponent',
+        'label' => '<i class="dashicons dashicons-controls-repeat"></i> Reusable',
```
steffenbew, line 19: The icon / text position looks too much to the top in the component dropdown. Any chance to fix that? Also would consider putting the icon at the end, and / or moving the component position to the very end of the list.

Bildschirmfoto 2020-09-29 um 20 12 07

Components/ReusableComponent/functions.php
koukouchris: Thats an easy change from acf field options, but the field is required so how a deselect would be helpful, just to clear the input? 
I will then need to trigger that event also in admin.js to clear link

Components/ReusableComponent/admin.js
koukouchris: I didn't know that thanks will look into it and refactor!

Components/ReusableComponent/functions.php
steffenbew: I didn't notice the field is required. Totally makes sense this way.

Components/ReusableComponent/_admin.scss
```diff
@@ -0,0 +1,12 @@
...
+.acf-fc-popup a {
+  &[data-layout='ReusableComponent'] {
+    align-items: center;
+    display: flex;
+
+    i {
+      margin-left: 5px;
+    }
+  }
+}
```
steffenbew, line 12: Looks very nice in the backend!

Components/ReusableComponent/admin.js
```diff
@@ -0,0 +1,22 @@
...
+    const oldPostId = $postLink.attr('data-postId')
+    const $hiddenEl = $fieldEl.find('[hidden]')
+    const href = $postLink.attr('href')
+
+    $hiddenEl.removeAttr('hidden')
+    $postLink.text(postTitle)
+    $postLink.attr('data-postId', postId)
+    $postLink.attr('href', href.replace(oldPostId, postId))
+    console.log(href)
+    console.log(postId)
```
steffenbew, line 20: Please remove the logs.

Components/ReusableComponent/admin.js
```diff
@@ -0,0 +1,22 @@
+/* globals acf */
+import $ from 'jquery'
+
+acf.addAction('select2_init', function ($select, args, settings, field) {
+  const $fieldEl = field.$el
+
+  $select.on('change', { $fieldEl: $fieldEl }, function () {
```
steffenbew, line 7: Doesn't this trigger on every select element in the backend? If so, it would be better to only fire on the specific select field of the reusable component.

Components/ReusableComponent/functions.php
```diff
@@ -0,0 +1,56 @@
...
+});
+
+function getACFLayout()
+{
+    return [
+        'name' => 'ReusableComponent',
+        'label' => 'Reusable <i class="dashicons dashicons-controls-repeat"></i>',
+        'sub_fields' => [
+            [
+                'label' => __('Select Component Set', 'flynt'),
```
steffenbew, line 22: Please see my following proposal:

Select Component Set => Select Reusable Components
Add or edit reusable components. => Add reusable components.
Edit component. => Move to same line.
Bildschirmfoto 2020-09-30 um 16 48 24

steffenbew (changes requested):

inc/fieldGroups/pageComponents.php
```diff
@@ -16,25 +16,31 @@
                 'button_label' => __('Add Component', 'flynt'),
                 'layouts' => [
                     Components\BlockCollapse\getACFLayout(),
-                    Components\BlockImage\getACFLayout(),
                     Components\BlockImageText\getACFLayout(),
+                    Components\BlockImage\getACFLayout(),
```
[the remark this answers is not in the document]
koukouchris: No, SortLines did it automatically

steffenbew:
@koukouchris Looking great!! Last thing I'm wondering about is, if the custom post type name should also be plural: `reusable-components`.

PR #311, opened 2020-10-01
Feat block anchor

koukouchris:


pixelsaft:
Hi @koukouchris, i have tested the block inside one project and found two issues:

1. speed: should also get min value (around 750) on small distances, because the scroll speed is a little bit to fast on small distances – in my opinion
2. NavigationBurger: Should be get closed before scroll starts

I’m fixed this two – really fast and ugly into a time intensive project. Hope the feedback helps you to improve the great block!

PR #320, opened 2020-10-16
feat(fallbackfeaturedimage): default / fallback featured image support

pixelsaft:
If the featured image is empty the fallback featured image will be used

### DIFF
--- a/inc/featuredImageFallback.php
+++ b/inc/featuredImageFallback.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Flynt\FeaturedImageFallback;
+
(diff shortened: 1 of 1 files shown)


steffenbew:
Thanks for your suggestion, @pixelsaft. Sorry it took me so long to respond. I discussed it with the core team. We came to the conclusion that we would like to keep this functionality out of the Flynt core, and rely on plugins like the following instead: https://de.wordpress.org/plugins/default-featured-image/
We didn't test the specific plugin yet, but if you have some input and experience with it or other ones, that would be great to learn from!

PR #321, opened 2020-10-16
feat(featureadminloginbranding): simple wp-admin branding

pixelsaft:
simple wp-admin login branding


steffenbew:
Thanks for your suggestion, @pixelsaft. Sorry it took me so long to respond. I discussed it with the core team. We came to the conclusion that we would like to keep this functionality out of the Flynt core, and rely on plugins like the following instead: https://logindesigner.com/

PR #340, opened 2020-11-30
feat(variables): add typography css variables with fallback

MadFlip:


harunbleech (approved):

steffenbew:
@MadFlip @harunbleech I'm not sure I understand the intentions here. Why introduce the fallbacks, if we're not using fallbacks for any other variables?

MadFlip:
> @MadFlip @harunbleech I'm not sure I understand the intentions here. Why introduce the fallbacks, if we're not using fallbacks for any other variables?

Because spacings and colors we always use in projects, but google fonts not, more often we use self-hosted fonts and in case you need to remove fonts option from customizer it easier to have fallback in one place rather then looking for all those variables in base style.

PR #343, opened 2020-12-14
feat(Asset): remove add functionality

domtra:


steffenbew (approved):
Love it! I'm all in for using the standards of WordPress in this case.

PR #369, opened 2021-03-22
(fix)baseStyle: updated deprecated function, preserve backwards compatibility

timohubois:


steffenbew:
Thanks for the contribution. I'm thinking if we should drop the backwards support. As a starter theme, I expect Flynt to be installed always with the newest version of WordPress.

Issue #430, opened 2022-03-17
Will you be interested in a "social sharing" component?


stefanobartoletti:
Hi, I'd like to contribute a new component, namely a "social sharing" icons block, something similar to what you can see here at the end of the article:

http://negozianti.sviluppo.host/news/2020/07/01/articolo-di-esempio-1/

Let me know :)

steffenbew:
Hi @stefanobartoletti, I think that could be a nice addition! 
We have a simple [social links](https://flyntwp.com/components/list-social/) component already, but I imagine a social sharing component would be visually integrated differently.

stefanobartoletti:
Thanks :)
Exactly, they will have a different visual integration since they also serve two different purposes.
I'll work on it and open a PR as soon as possible.

timohubois:
See #437.

PR #437, opened 2022-04-04
feat: add new social share component

stefanobartoletti:
This PR is about #430 

This component simply adds some social sharing buttons.

Some considerations:

- I have styled the buttons as few as possible, to make the component be more consistent with the overall look of the theme.
- Only a few social networks are included, mainly because the Feather icon theme seems to support only a few brand icons.
- Social sharing URLs and their parameters were taken from this [documentation](https://github.com/bradvin/social-share-urls)
- Social network colors were taken from https://brandcolors.net/

Please let me know if this PR is fine as it is or if you think that something should be modified :)


timohubois:
@stefanobartoletti first: Thank you for the great work and the PR. Currently, we are working on the next version of Flynt and we think, that we should only provide a as less as possible amount of components from the core, to focus more on a boilerplate approach and for now we need to close this PR.

**But we are also internally thinking about a more elegant way to share custom made components. What would you like to see? And what would you expect from this feature?**

I am very much looking forward to your feedback!

stefanobartoletti:
I understand, this is not a problem :)

I'm not currently using Flynt, when working on Wordpress I mainly use a theme of my own creation, https://github.com/stefanobartoletti/bricks, I found on Flynt some interesting ideas and a component concept that was also what I had in mind for Bricks. 

Anyway let me know if you need some other kind of feedback.

PR #439, opened 2022-05-02
fix(webpackConfig): add content hash in chunked css file name

prathamesh-gharat:
Required for dynamically importing CSS files.

Solves error: `Conflict: Multiple chunks emit assets to the same filename assets/main.css (chunks 2 and main)`.


prathamesh-gharat:
Causes other issues. Closing for now. Will try to provide additional details later.

PR #483, opened 2023-04-12
Load the minified version of jquery

prathamesh-gharat:
I noticed this while running a Lighthouse test.


timohubois:
@prathamesh-gharat thanks for the PR! We released a new version of Flynt where we the fix is not needed anymore.

PR #494, opened 2023-07-18
Fix curly brace usage in string

szepeviktor:
If PHP it is `{$var}` not `${var}`.


timohubois:
@szepeviktor thanks for this PR!

szepeviktor:
Glad to contribute.

PR #495, opened 2023-07-18
Stop exporting development files

szepeviktor:
There is how to test what gets into GitHub ZIP files.

```bash
git archive HEAD | tar --list
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+/.editorconfig export-ignore
+/.github/ export-ignore
+/.gitignore export-ignore
+/.scrutinizr.yml export-ignore
+/.travis.yml export-ignore
+/composer.lock export-ignore
+/package-lock.json export-ignore
+/phpcs.xml.dist export-ignore
+/vite-plugin-flynt.js export-ignore
+/vite.config.js export-ignore


timohubois (approved):

aaronmeder:
@timohubois @szepeviktor 

I noticed this change when checking out the wonderful Flynt v2 today (I downloaded the zip 😉).
Why is this a desired behaviour? Would like to understand.

Since Flynt is a developer starter theme, not a theme to use right away without dev work it seems more suitable to me to include the files - what do you guys think? Not having files like the vite config available seems a bit counter-intuitive to me.

Thanks for clarifying.
Aaron

szepeviktor:
... I suspect that you develop without git - that is something I @timohubois will answer.

aaronmeder:
> @aaronmeder It is very simple.
> 
> When you develop you issue the `git clone` command and start committing. `export-ignore` does not influence that.
> 
> When you add Flynt as a dependency you do not need development files in `vendor/`.
> 
> I think that is all.

Hey @szepeviktor yep it's clear to me that working with `Git clone` is not impacted. That's the way I usually work when creating a new theme based on Flynt. 
Sometimes I go and download a zip of the repo to look at the files in my code editor - which was what I was doing to get a first look at Flynt 2 but then realized the ZIP was missing some files.

I however don't get what you mean by "When you add Flynt as a dependency" - how would that work and in which situation would someone do that? As far as I know Flynt doesn't support a parent/child theme setup. Thanks for helping me understand.

PR #496, opened 2023-07-19
Upgrade Scrutinizr OS

szepeviktor:
Nodejs v18 needs modern libc.
https://scrutinizer-ci.com/docs/guides/upgrading-trusty-image


timohubois:
@domtra can you take a look on it?

szepeviktor:
@timohubois Turns out the file name had a typo! `.scrutinizer.yml`

szepeviktor:
> node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)

Still older libs than necessary. Upgrading OS ...

szepeviktor:
I give up.

domtra:
@szepeviktor thanks for your effort.

It seems to be impossible to get PHP 8.0 and node 18.x running under any of the scrutinizer images. I might try it with a custom image or use PHP 8.1 instead of 8.0.

When this was working, it gave us a simple guideline with a code quality score from 0-10. This was very convenient. However, the service seems to be a bit outdated and it is not quite active. I have tried some other services yesterday, mainly the ones that have github app integration and are recommended. all of them 'worked', but i did not like the results that came out of the box.

I know that you are big on static analysis, but honestly, i (currently) do not want to add extra dependencies just for static analysis. I would rather have a service like scrutinizer with one rating that takes multiple issues into account. Can you recommend any service that has free tiers for open source?

I really appreciate your quest for high coding standards. Thanks again!

szepeviktor:
I could help keeping potential problems/bug/tech debt az Zero.

PR #501, opened 2023-08-04
fix(FeatureFlexibleContentExtension): add component tooltip positioning

aaronmeder:
Resolves issue mentioned here for me: https://github.com/flyntwp/flynt/discussions/499
Works locally in Firefox and Chrome.

Previously the tooltip could be positioned out of view when there's a scrollbar: 
Bildschirmfoto 2023-08-04 um 10 28 11

This is most likely my first direct code commit to an open source project through Github btw...hope everything  is fine in terms of the way how it's done 😉 Changing one character seems like a good start 😛


timohubois:
@aaronmeder we merged your PR – many thanks for the fix!

PR #504, opened 2023-08-09
Iconify API support

thomasnavarro:
Hi,

I use Iconify in my projects for icon management, and it would be great if it were included in Flynt because it's really convenient and their API is excellent.

In addition to accessing the vast icon library, I've created a Vite plugin to integrate custom icons. You simply need to drag and drop an SVG icon into the folder and call it using the Iconify web component.

For example:
- I add the `anesthesia.svg` file to the `assets/icons` folder,
- The `interlude.json` file updates automatically,
- I can now display the icon in my theme using `<iconify-icon icon="interlude:anesthesia"></iconify-icon>`.

To access the available icons: https://icones.js.org/
To learn more about Iconify: https://iconify.design/


timohubois:
@thomasnavarro Thanks for the PR. 

We will discuss the feature at our next team meeting and come back with a feedback ✌️

timohubois:
@thomasnavarro many thanks for the great idea, the PR, to start a [discussion](https://github.com/flyntwp/flynt/discussions/505) and for your dedication! 

We took a look at the changes and discussed the feature in our team. The suggested approach includes some adjustments – as well as a new custom plugin for vite and other new files or adjustments at existing files like at the assets.php. 

Flynt is a boilerplate, and we don't want to dictate any particular directions, and we always try to keep complexity and dependencies to a minimum. 

But there is another package in the wild, which icons can probably implement with less adjustments: [unplugin-icons](https://github.com/antfu/unplugin-icons)

After careful consideration and review, we regret that we will not be able to accept your pull request at this time. Please know that this decision was not taken lightly, and we highly value your input. 

Feel free to reach out with new ideas or modifications and we would be more than happy to see a new Pull Requests in the future 🚀

PR #551, opened 2024-04-23
Fixing use of `execCommand('copy')`

gRoberts84:
When using the Copy Link button on the Anchor component without `navigator.clipboard` support, it reverts to `execCommand('copy')` which requires the content to copy to be selected.

This PR adds code to ensure the link is selected before calling `execCommand('copy')`.

### DIFF
--- a/Components/BlockAnchor/admin.js
+++ b/Components/BlockAnchor/admin.js
@@ -47,6 +47,13 @@
       const $copyMessage = $anchorField.find('.anchorLink-message')
 
       if (!navigator.clipboard) {
+        const selection = window.getSelection()
+        if (selection.rangeCount > 0) {
+          selection.removeAllRanges()
+        }
+        const range = document.createRange()
+        range.selectNode($anchorLinkInput.get(0))
+        selection.addRange(range)
         document.execCommand('copy')
       } else {
         navigator.clipboard.writeText($anchorLinkInput.text()).then(


timohubois:
@gRoberts84 many thanks for the pr. I took a short to can I use: https://caniuse.com/?search=navigator.clipboard

It seems that `navigator.clipboard` has widely support (95.37%; [04-26-2024]) which is pretty close to the the `browserlist` setting supports [array-flat](https://caniuse.com/?search=array%20flat) (96.82%; [04-26-2024]) inside the [package.json](https://github.com/flyntwp/flynt/blob/a2f5ac6f4d12c053e26f8d7fb8550183ef64c469/package.json#L137). 

Did you run in trouble with the current implementation at a specific browser? I am looking forward to learning more details.

gRoberts84:
Hi @timohubois 

Apologies for the delay in getting back to you.

I had mistaken it only working over HTTPS as a bug but it had highlighted an issue where if `navigator.clipboard` isn't available for any reason (e.g. insecure local environment) then it tries to use `execCommand('copy')` without any text being selected, so it copies nothing.

Appreciate that under most circumstances, this won't be an issue but someone could come across it again on an insecure site.

gRoberts84:
@timohubois appreciate your view on this and happy to close the PR.

PR #552, opened 2024-04-28
fix(CleanHead): move emoji detection script removal from 'init' to 'admin_init'

kunsang:
If you use utf-8 emojis inside wp-admin, the `'print_emoji_detection_script'` script was still running, and the admin page loads were slowed down due to 404s from `cdn.jsdelivr.net`.

Screenshot 2024-04-28 at 21 43 05

Moving the actions from `'init'` to `'admin_init'` fixes this.

Screenshot 2024-04-28 at 21 56 00


timohubois:
@kunsang many thanks for the pr. Can you please provide some instructions how to reproduce the issue as seen in the screenshots?

PR #609, opened 2025-08-18
fix(FeatureFlexibleContentExtension): resolve compatibility issues with ACF Pro 6.5.0

timohubois:


timohubois:
@aaronmeder and @harunbleech made some minor style improvements: adjusted the spacing inside the pop-up for better readability and ensured that the "no results" state no longer causes layout jumps. Screenshots now overlap again.

For the new ACF rename feature, only the renamed title is displayed. This feels more natural and closer to the previous behavior, even though the implementation is a bit hacky and doesn’t fully uses the native ACF behavior.

Let me know what you think!

TheChriZZoW:
@timohubois Yes, they work for me. But I don't think the improvements were really necessary.

PR #623, opened 2026-04-07
feat(assets): enqueue theme stylesheet

mahmoud-alawad:
Previously, the main theme CSS was not enqueued by default, requiring manual inclusion via assets.php. I have now enabled default enqueuing for this file.
This allows for quick CSS overrides and emergency production fixes directly through the stylesheet, bypassing the need for a full theme rebuild when immediate changes are required.


timohubois:
@mahmoud-alawad Thanks for the PR! I understand the use case, but I don’t think this file should be enqueued by default.

Since it’s empty by default, enqueuing it would trigger an unnecessary request for an empty file. Additionally, always loading it could make it less clear for new developers where custom CSS should actually be added.

